PR #3, opened 2015-12-06
Changes to allow bring up of bare metal setup and aci interface

jojimt:
Summary:
- Added two new host groups
- Added some custom facts 
- Templated netplugin args to pick up the correct vlan_if
- Fixed some misc errors related to clean up etc.
- Added an aci role to handle aci specifc tasks

### DIFF
--- a/roles/aci/tasks/main.yml
+++ b/roles/aci/tasks/main.yml
@@ -1,0 +1,23 @@
+---
+# This role contains tasks specific to aci
+
+  - name: Install docker-py as a workaround for Ansible issue
+    pip: name=docker-py version=1.2.3
+    when: run_as == "master"
+
+  - name: start aci-gw container
+    docker:
+        name: contiv-aci-gw
+        image: contiv/aci-gw
+        state: reloaded
+        pull: always
+        net: host
+        env:
+             APIC_URL: "{{ apic_url }}"
+             APIC_USERNAME: "{{ apic_username }}"
+             APIC_PASSWORD: "{{ apic_password }}"
+             APIC_LEAF_NODES: "{{ apic_leaf_nodes }}"
+    when: run_as == "master"
+    ignore_errors: yes
+
+
--- a/roles/base/tasks/main.yml
+++ b/roles/base/tasks/main.yml
@@ -1,6 +1,27 @@
 ---
 # This role contains tasks for install base packages
 
+- name: ensure custom facts directory exists
+  file: >
+    path=/etc/ansible/facts.d
+    recurse=yes
+    state=directory
+
+- name: install custom fact module for docker
+  template: >
+    src=docker_facts.j2
+    dest=/etc/ansible/facts.d/docker_facts.fact
+    mode=0755
+
+- name: install custom fact module for swarm
+  template: >
+    src=swarm_facts.j2
+    dest=/etc/ansible/facts.d/swarm_facts.fact
+    mode=0755
+
+- name: reload ansible local facts
+  setup: filter=ansible_local
+
 - name: install openstack kilo repo
   yum: "name=https://repos.fedorapeople.org/repos/openstack/openstack-kilo/rdo-release-kilo-1.noarch.rpm"
   when: ansible_os_family == "RedHat"
@@ -18,6 +39,7 @@
     - librbd-dev
     - lshw
   when: ansible_os_family == "Debian"
+  ignore_errors: yes
 - name: install base packages (redhat)
   yum: "update_cache=yes name={{ item }}"
   with_items:
@@ -54,16 +76,24 @@
   copy:
     dest: /etc/profile.d/00golang.sh
     content: "export PATH=/opt/golang/bin:/usr/local/go/bin:$PATH\nexport GOPATH=/opt/golang"
+- name: remove older docker
+  file: path=/usr/bin/docker state=absent
+  when: ("{{ansible_local.docker_facts.local_docker_version}}" < "1.9.0") or ("{{ansible_local.docker_facts.local_experimental}}" == "true")
 - name: install docker
   shell: creates=/usr/bin/docker curl https://get.docker.com | bash
+  register: docker_install_result
 - name: add vagrant user to docker group
   user: name=vagrant groups=docker append=yes
+- name: remove older swarm
+  file: path=/usr/bin/swarm state=absent
+  when: ("{{ansible_local.swarm_facts.local_swarm_version}}" < "1.0.0") or ("{{ansible_local.swarm_facts.local_experimental}}" == "true")
 - name: download and install swarm
   get_url:
     validate_certs: no
     url: https://cisco.box.com/shared/static/0txiq5h7282hraujk09eleoevptd5jpl
     dest: /usr/bin/swarm
     mode: u=rwx,g=rx,o=rx
+  register: swarm_install_result
 - name: download ovs binaries (debian)
   get_url:
     validate_certs: no
--- a/roles/base/tasks/main.yml
+++ b/roles/base/tasks/main.yml
@@ -25,6 +25,9 @@
 - name: install openstack kilo repo
   yum: "name=https://repos.fedorapeople.org/repos/openstack/openstack-kilo/rdo-release-kilo-1.noarch.rpm"
   when: ansible_os_family == "RedHat"
+- name: apply workaround for apt get failure
+  shell: rm -rf /var/lib/apt/lists/*
+  when: ansible_os_family == "Debian"
 - name: install base packages (debian)
   apt: "update_cache=yes name={{ item }}"
   with_items:
@@ -39,7 +42,6 @@
     - librbd-dev
     - lshw
   when: ansible_os_family == "Debian"
-  ignore_errors: yes
 - name: install base packages (redhat)
   yum: "update_cache=yes name={{ item }}"
   with_items:
--- a/roles/base/templates/docker_facts.j2
+++ b/roles/base/templates/docker_facts.j2
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+dver=$(sudo docker version | awk '/Version/{print $2; exit 0}')
+exper=$(sudo docker version | awk '/Experiment/{print $2; exit 0}')
+if [ -z "$dver" ]; then
+  dver="0.0"
+fi
+
+if [ -z "$exper" ]; then
+  exper="false"
+fi
+
+cat <<EOF
+
+{
+  "local_docker_version" : "$dver",
+  "local_experimental"   : "$exper"
+}
+
+EOF
--- a/roles/contiv_network/tasks/main.yml
+++ b/roles/contiv_network/tasks/main.yml
@@ -1,10 +1,25 @@
 ---
 # This role contains tasks for configuring and starting netmaster and netplugin service
 
+- name: remove older netmaster
+  file: path=/usr/bin/netmaster state=absent
+
+- name: remove older netplugin
+  file: path=/usr/bin/netplugin state=absent
+
+- name: remove older compose
+  file: path=/usr/bin/docker-compose state=absent
+
+- name: remove older contivctl
+  file: path=/usr/bin/contivctl state=absent
+
+- name: remove older tar file
+  file: path=/tmp/contivnet.tar.bz2 state=absent
+
 - name: download netmaster and netplugin
   get_url:
     validate_certs: no
-    url: https://github.com/contiv/netplugin/releases/download/v0.0.0-11-12-2015.00-55-50.UTC/netplugin-v0.0.0-11-12-2015.00-55-50.UTC.tar.bz2
+    url: https://cisco.box.com/shared/static/hs2gp4jfvssr1bu2geh3o6yipse9ljjd.bz2
     dest: /tmp/contivnet.tar.bz2
 
 - name: install netmaster and netplugin
@@ -21,33 +36,33 @@
   copy: src=netmaster.service dest=/etc/systemd/system/netmaster.service
   when: run_as == "master"
 
-- name: start netmaster
-  shell: systemctl daemon-reload && systemctl start netmaster
+- name: restart netmaster
+  shell: systemctl daemon-reload && systemctl stop netmaster && systemctl start netmaster
   when: run_as == "master"
 
 - name: copy environment file for netplugin
-  copy: src=netplugin dest=/etc/default/netplugin
+  template: src=netplugin.j2 dest=/etc/default/netplugin mode=0644
 
 - name: copy systemd units for netplugin
   copy: src=netplugin.service dest=/etc/systemd/system/netplugin.service
 
-- name: start netplugin
-  shell: systemctl daemon-reload && systemctl start netplugin
+- name: restart netplugin
+  shell: systemctl daemon-reload && systemctl stop netplugin && systemctl start netplugin
 
 - name: setup netmaster host alias
   shell: echo "{{ service_vip }} netmaster" >> /etc/hosts
 
 # XXX: need to move the following to correct roles
-- name: install contivctl
-  shell: source /etc/profile.d/00golang.sh && go get github.com/contiv/contivctl
-
-- name: install docker-compose
-  shell: pip install docker-compose
-
-- name: install contiv-compose
-  shell: >
-     source /etc/profile.d/00golang.sh && \
-     go get github.com/contiv/libcompose && \
-     cd $GOPATH/src/github.com/contiv/libcompose && \
-     make binary && \
-     ln -s `pwd`/bundles/libcompose-cli_linux-amd64 $GOPATH/bin/contiv-compose
+#- name: install contivctl
+#  shell: source /etc/profile.d/00golang.sh && go get github.com/contiv/contivctl
+#
+#- name: install docker-compose
+#  shell: pip install docker-compose
+#
+#- name: install contiv-compose
+#  shell: >
+#     source /etc/profile.d/00golang.sh && \
+#     go get github.com/contiv/libcompose && \
+#     cd $GOPATH/src/github.com/contiv/libcompose && \
+#     make binary && \
+#     ln -s `pwd`/bundles/libcompose-cli_linux-amd64 $GOPATH/bin/contiv-compose
--- a/roles/contiv_network/tasks/main.yml
+++ b/roles/contiv_network/tasks/main.yml
@@ -19,7 +19,7 @@
 - name: download netmaster and netplugin
   get_url:
     validate_certs: no
-    url: https://cisco.box.com/shared/static/hs2gp4jfvssr1bu2geh3o6yipse9ljjd.bz2
+    url: "{{ contivnet_bin_url }}"
     dest: /tmp/contivnet.tar.bz2
 
 - name: install netmaster and netplugin
@@ -51,18 +51,3 @@
 
 - name: setup netmaster host alias
   shell: echo "{{ service_vip }} netmaster" >> /etc/hosts
-
-# XXX: need to move the following to correct roles
-#- name: install contivctl
-#  shell: source /etc/profile.d/00golang.sh && go get github.com/contiv/contivctl
-#
-#- name: install docker-compose
-#  shell: pip install docker-compose
-#
-#- name: install contiv-compose
-#  shell: >
-#     source /etc/profile.d/00golang.sh && \
-#     go get github.com/contiv/libcompose && \
-#     cd $GOPATH/src/github.com/contiv/libcompose && \
-#     make binary && \
-#     ln -s `pwd`/bundles/libcompose-cli_linux-amd64 $GOPATH/bin/contiv-compose
--- a/roles/contiv_network/templates/netplugin.j2
+++ b/roles/contiv_network/templates/netplugin.j2
@@ -1,0 +1,7 @@
+{# If an interface is specified for netplugin, use it #}
+{# Otherwise, hope for the best                       #}
+{% if netplugin_if is defined %}
+NETPLUGIN_ARGS='--native-integration=true -vlan-if {{netplugin_if}}'
+{% else %}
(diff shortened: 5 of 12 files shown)


jainvipin:
@joji - any reason this is only for `demo`?  I mean why not use it for the real setup too?

roles/aci/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+# This role contains tasks specific to aci
+
+  - name: Install docker-py as a workaround for Ansible issue
```
jainvipin, line 4: I am curious what is the ansible issue: is it to allow `image` and other tags to be specified in ansible?

roles/base/tasks/main.yml
```diff
@@ -18,6 +39,7 @@
     - librbd-dev
     - lshw
   when: ansible_os_family == "Debian"
+  ignore_errors: yes
```
jainvipin: Why do we we ignore errors?

erikh: yeah, we absolutely cannot ignore errors here

roles/base/templates/docker_facts.j2
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+dver=$(sudo docker version | awk '/Version/{print $2; exit 0}')
+exper=$(sudo docker version | awk '/Experiment/{print $2; exit 0}')
```
erikh, line 4: can't we just do this for the tasks in teh ansible above, so we can get rid of the docker dependency which seems to add a lot of extra stuff?

roles/contiv_network/tasks/main.yml
```diff
@@ -1,10 +1,25 @@
...
+  file: path=/usr/bin/contivctl state=absent
+
+- name: remove older tar file
+  file: path=/tmp/contivnet.tar.bz2 state=absent
+
 - name: download netmaster and netplugin
   get_url:
     validate_certs: no
-    url: https://github.com/contiv/netplugin/releases/download/v0.0.0-11-12-2015.00-55-50.UTC/netplugin-v0.0.0-11-12-2015.00-55-50.UTC.tar.bz2
+    url: https://cisco.box.com/shared/static/hs2gp4jfvssr1bu2geh3o6yipse9ljjd.bz2
```
erikh: find a way to parameterize this; cluster is relying on the releases to be in there, so if you need a different url let's make it a var you can pass into ansible.

roles/contiv_network/tasks/main.yml
```diff
@@ -21,33 +36,33 @@
...
-  shell: pip install docker-compose
-
-- name: install contiv-compose
-  shell: >
-     source /etc/profile.d/00golang.sh && \
-     go get github.com/contiv/libcompose && \
-     cd $GOPATH/src/github.com/contiv/libcompose && \
-     make binary && \
-     ln -s `pwd`/bundles/libcompose-cli_linux-amd64 $GOPATH/bin/contiv-compose
+#- name: install contivctl
```
erikh: if we're going to comment this code out, I would prefer we remove it. We can get it from source control later.

fwiw @mapuri, @jainvipin building and installing the contiv-compose tool takes enough time for ansible to timeout.

We could use the ansible async stuff but this is the best answer I could find regarding a general timeout: http://grokbase.com/t/gg/ansible-project/149fts2xd2/how-to-set-timeout-for-shell-command which is not very promising. 

What I think we need to do is improve build performance before we can install this. I think we can come up with some good ways to do this such as building images with godeps preinstalled, or inventing release tooling through travis or whatnot for contiv-compose.

So my suggestion here is that we just remove this code for now and restore it later.

erikh:
Rest of it LGTM!

erikh:
Oh. Please rewrite your commit to have your email and real name. Thanks.

roles/contiv_network/tasks/main.yml
jainvipin: @erikh, @joji, @mapuri:
must we build? Can we move to released versioning here?

erikh: that's another option, yes, that I would prefer.

site.yml
```diff
@@ -60,3 +60,30 @@
...
   - { role: contiv_network, run_as: worker }
   - { role: contiv_storage, run_as: worker }
+
+# net-master hosts correspond to cluster machines that run the master
+# logic of the contiv net services
+- hosts: net-master
+  sudo: true
+  environment: env
+  roles:
+  - { role: base }
```
mapuri, line 72: consider putting bare-metal host in both the `devtest` group and `net-master` group in `inventory` file. This way this host-group can be used for vagrant setups as well, where the `base` role has already been run and it will save some provisioning time.

roles/aci/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+# This role contains tasks specific to aci
```
mapuri, line 2: I had described this in our call earlier but I might have not been clear. We should move the aci role as task inside `contiv_network` role which is used for setting up the networking service. The aci specific tasks can be trigger based deploymnent-environment. If you wish to keep aci specific tasks in a separate file, you can add a tasks file in `contiv_network` role itself and include it in main `contiv_network` role. For instance, checkout how it is done here: https://github.com/contiv/ansible/blob/master/roles/ceph-common/tasks/main.yml

roles/base/tasks/main.yml
```diff
@@ -54,16 +76,24 @@
   copy:
     dest: /etc/profile.d/00golang.sh
     content: "export PATH=/opt/golang/bin:/usr/local/go/bin:$PATH\nexport GOPATH=/opt/golang"
+- name: remove older docker
```
mapuri, line 81: This is minor but I think instead of a task to remove docker, we should enhance/fix the task to install docker to do the right thing i.e. only install/upgrade docker if we are at version lower that 1.9? This would help us have all docker version specific logic in one task.

roles/base/templates/docker_facts.j2
mapuri, line 4: +1, same is true for swarm related facts as well.

roles/contiv_network/tasks/main.yml
```diff
@@ -1,10 +1,25 @@
 ---
 # This role contains tasks for configuring and starting netmaster and netplugin service
 
+- name: remove older netmaster
```
mapuri, line 4: This can impact ansible runtimes if same playbook needs to be played twice.

We can instead during create steps check if we are at a desired version or not and take appropriate action.

roles/contiv_network/tasks/main.yml
mapuri: yeah, that is correct. This will break cluster. Passing in a url or a binary location as a var is a better way.

roles/contiv_network/tasks/main.yml
```diff
@@ -21,33 +36,33 @@
   copy: src=netmaster.service dest=/etc/systemd/system/netmaster.service
   when: run_as == "master"
 
-- name: start netmaster
-  shell: systemctl daemon-reload && systemctl start netmaster
+- name: restart netmaster
```
mapuri, line 39: we can use the `service` module here, that will simplify this task.

Also I am wondering if always restarting netmaster is desirable in a production setup. May be this task needs to restart only if a new netmaster binary was copied?

roles/contiv_network/tasks/main.yml
```diff
@@ -21,33 +36,33 @@
...
 - name: copy environment file for netplugin
-  copy: src=netplugin dest=/etc/default/netplugin
+  template: src=netplugin.j2 dest=/etc/default/netplugin mode=0644
 
 - name: copy systemd units for netplugin
   copy: src=netplugin.service dest=/etc/systemd/system/netplugin.service
 
-- name: start netplugin
-  shell: systemctl daemon-reload && systemctl start netplugin
+- name: restart netplugin
```
mapuri, line 49: same a above we can use `service` module to start/restart a service.

roles/contiv_network/tasks/main.yml
mapuri: +1 to remove this stuff if it is not needed. This was added as a quick solution to get dockercon demo setup going.

In future, if we need `contiv-compose` again we shall first introduce released versions for it and use them instead.

roles/contiv_network/templates/netplugin.j2
```diff
@@ -0,0 +1,7 @@
+{# If an interface is specified for netplugin, use it #}
+{# Otherwise, hope for the best                       #}
```
mapuri, line 2: It might be better to fail ansible instead of leaving the setup in some undesired state?

roles/docker/tasks/main.yml
```diff
@@ -22,4 +22,6 @@
 
 - name: restart docker
   shell: systemctl daemon-reload && systemctl stop docker && systemctl start docker-tcp.socket && systemctl start docker
-  when: docker_env_conf_result.changed or docker_tcp_conf_result.changed or docker_svc_conf_result.changed
+  when: docker_install_result.changed or swarm_install_result.changed or docker_env_conf_result.changed or
```
mapuri: I think `docker_install_result` and `swarm_install_result` variable will not be available if `base` role is not run, in which case this will fail.

roles/docker/tasks/main.yml
```diff
@@ -22,4 +22,6 @@
...
+        docker_tcp_conf_result.changed or docker_svc_conf_result.changed or force_restart
```
mapuri: what is the condition for `force_restart`? if it is just expressing user's desire then may be we should let user manaually bump the service, assuming they know what they are doing?

roles/etcd/tasks/main.yml
```diff
@@ -1,6 +1,8 @@
 ---
 # This role contains tasks for configuring and starting etcd service
 
+#- name: debug
```
mapuri: suggest to cleanup commented code.

mapuri:
thanks for the changes @jojimt . They are in right directions and I think we shall be able to wriggle out any leftover comments in a few iterations.

roles/aci/tasks/main.yml
jojimt, line 4: The issue is that docker-py is needed for docker. Without it, docker module returns random errors. Found this from the github discussion on the topic.
https://github.com/ansible/ansible-modules-core/issues/1792

roles/base/tasks/main.yml
jojimt: I don't like to ignore errors. Unfortunately, this task intermittently returns apt failures. If anyone knows a fix/workaround for that, we can add that and remove the ignore errors.

roles/base/templates/docker_facts.j2
jojimt, line 4: The issue I am solving is this: If we already have a docker installed, ansible does not install the new version. This is the solution I came up with after some digging. If you know a better way, please suggest it and I will change accordingly.

roles/contiv_network/tasks/main.yml
jojimt: Again, this is not my preference, rather something forced by the circumstances :) The release folder does not point to the right binaries. IOW, the binaries in the folder pointed do not work:( Therefore, I created a folder on box with known working binaries and used that. Now, I can definitely parameterize this temporary work-around and I will do just that!

roles/contiv_network/tasks/main.yml
jojimt: OK, I will remove it.

site.yml
jojimt, line 72: As this is more of an optimization, I would prefer doing this as a follow incremental change.

roles/aci/tasks/main.yml
jojimt, line 2: This is really a matter of taste, analogous to breaking up of code modules/api's etc. Based on the things we do for aci, it seems to make sense to have a separate task. At least I can't see how that would hurt...

roles/base/tasks/main.yml
jojimt, line 81: Can you provide specifics of how this can be achieved? If so, I can implement it, perhaps as a follow on.

roles/contiv_network/tasks/main.yml
jojimt, line 4: Unfortunately, contivnet binaries do not support version checks now. Therefore, I cannot implement this using facts they way I did for docker and swarm. But, this can definitely be done in the future.

roles/contiv_network/tasks/main.yml
jojimt, line 39: This is existing code. We can definitely enhance it in future.

roles/contiv_network/templates/netplugin.j2
jojimt, line 2: The existing scripts do not pass this argument. I did not want to break existing stuff. That's why I did not bail out here.

roles/docker/tasks/main.yml
jojimt: True. But, docker role should not be run without running base role.

roles/docker/tasks/main.yml
jojimt: Yes, this will be set when the user wants to force a restart regardless of other conditions... e.g. when trying to fix a broken set up.

roles/etcd/tasks/main.yml
jojimt: will do that.

roles/base/tasks/main.yml
erikh: This task is absolutely critical; it installs the dependencies needed by 
several parts of our project.

If it's erroring consistently, we can probably work around that, we just 
need to know what's happening.

-Erik

roles/aci/tasks/main.yml
erikh, line 4: Right. What I'm saying is all we do is extract the version information 
for installing the latest version of docker.

If we can do that with a shell command, a lot of this no longer becomes 
necessary. Can we do that instead, or is there something I'm missing 
here?

roles/base/templates/docker_facts.j2
erikh, line 4: I'm guessing something like this will work: https://gist.github.com/erikh/61eb4216c17cb842c4f7

jojimt, line 4: With some changes to cover docker not being present etc., this can be made to work. But, how is it better than what I implemented? I prefer my method because it is easier to extend.

roles/base/tasks/main.yml
jojimt: Below is the error. Let me know if you any clues.
failed: [node1] => (item=unzip,vim-nox,curl,python-software-properties,git,mercurial,build-essential,perl,librbd-dev,lshw) => {"failed": true, "item": "unzip,vim-nox,curl,python-software-properties,git,mercurial,build-essential,perl,librbd-dev,lshw"}
msg: Could not fetch updated apt files

roles/base/templates/docker_facts.j2
erikh, line 4: My only argument here is that we don't really need it to do anything we 
can't trivially do on the command line. If we expect a lot of change 
around this soon, then I will leave it alone, but the code 
footprint/orchestration is significant and I think we should measure 
carefully.

jojimt:
I guess we could use it for real setup too. I did not want claim it because we have not tested it in a real deployment.

jainvipin:
Let's target real setup (if you can also change the title also), where demo happens to be a smaller setup use case for this.

> I guess we could use it for real setup too. I did not want claim it because we have not tested it in a real deployment.

site.yml
mapuri, line 72: actually it's not an optimization. the `base` role is expected to be run just once (for instance, for configuring our base os image that we will ship) and this how i would expect it to be in production as well i.e. os image comes up configured with already run and then we use other roles to provision our services on top.

roles/aci/tasks/main.yml
mapuri, line 2: sorry but I am still -1 for this.

this is more of organization issue and breaks consistency with rest of the roles organization. right now `contiv_network` is the role for configuring contiv networking which shall include both aci and standalone deployments.

roles/base/tasks/main.yml
mapuri, line 81: sure, I was thinking to rewrite the `docker install` task as follows. This is just a sample to show the idea and might have bugs. We can also wrap all of this is in a single docker install script as well and replace multiple tasks with one. The benefit is that say in future if docker really start providing a way to pull specific versions, we can just replace our custom script with the docker's script:

```
- stat: path=/usr/bin/docker
  register: docker_bin

- shell:  docker version | awk '/Version/{print $2; exit 0}' 
  register: docker_version

- name: install docker
  shell: curl https://get.docker.com | bash
  when: docker_bin.exists == False or docker_ver.output < 1.9
```

roles/contiv_network/tasks/main.yml
mapuri, line 4: how about a m5sum? But I agree this would easier if our binaries support a version flag.

I would still avoid these remove task as a workaround for some missing function,as I fear we will start depending on it soon and it will be hard to remove later.

What is the usecase you are trying to solve by removing the binaries? Can we use the fact that a new binary was indeed downloaded by looking at the result of the `get_url` result?

roles/contiv_network/tasks/main.yml
mapuri, line 39: Image of Yaktocat

Also it's not a big change. But I can accomodate this if you would like to keep it this way.

roles/contiv_network/templates/netplugin.j2
mapuri, line 2: We should fix existing stuff than leaving it broken as the code has to work correctly in all conditions. If stuff fails, it is an indicator of the fact that those places need to be fixed.

Can we fix this by defining a `netplugin_if` variable (in contiv_network specific vars files) and assigning it a default value like we do for `monitor_interface` value?

roles/docker/tasks/main.yml
mapuri: well `base` is meant to be run just once (when the os image is built)..so `docker` role is run on an image configured from `base` but it might not be in same host-group in which case this will fail.

I think `base` role itself needs to be cleaned up to not install docker and move it to `docker` role but unless you are taking care of that this will break existing suff.

roles/docker/tasks/main.yml
mapuri: ok, in that case let's not add it to ansible.

site.yml
jojimt, line 72: I think I mis-understood your comment. Regardless, [devtest] has roles that I do not want to run (e.g. ceph). Additionally, base role will be run only once regardless of whether I add the host to [devtest] in inventory or not.

roles/aci/tasks/main.yml
jojimt, line 2: I think we have to figure out a way forward because I don't agree that it breaks consistency.

roles/contiv_network/tasks/main.yml
jojimt, line 4: I am planning to add versioning to contivent binaries and then change this.

roles/contiv_network/templates/netplugin.j2
jojimt, line 2: While I agree with fixing broken things, I don't think it needs to be part of this effort. I believe in fixing things incrementally.

roles/docker/tasks/main.yml
jojimt: OK - I will remove this conditional.

roles/docker/tasks/main.yml
jojimt: Can you please explain why.

roles/contiv_network/tasks/main.yml
jojimt, line 39: That is an interesting slogan :)

site.yml
mapuri, line 72: you missed the point. To put it the other way around, if you use `net-master` host-group for provisioning a vagrant environment it would run `base` role twice, right? 

The `devtest` role is used for setting up basic dev environment on bare-metals (by `contiv/lab`) and for vagrant box images (by `contiv/build`). So assuming, the new host-groups are built on top of such machines then `base` should not be included again in any of the host-groups that provision things on top of it otherwise we are unnecessary running same ansible twice. Again this is not an optimization but more of how things are setup right now.

roles/aci/tasks/main.yml
mapuri, line 2: sure, what part you don't agree with my described approach?

roles/contiv_network/tasks/main.yml
mapuri, line 4: cool, thanks!

roles/contiv_network/templates/netplugin.j2
mapuri, line 2: hmmm, unfortunately I don't agree. This code breaks certain things, how is fixing it incrementally right thing to do?

roles/docker/tasks/main.yml
mapuri: ansible is a way to configure things and upgrade stuff just like any configuration management code. So, usually re-running same playbook should not have any side-effects if nothing changed. Adding convenience variable for a user overload this logic and make it complex to maintain.

site.yml
jojimt, line 72: "To put it the other way around, if you use net-master host-group for provisioning a vagrant environment it would run base role twice, right?"

It is not intent of this PR to use net-master host group for provisioning vagrant. However, it can be the intent of another PR.

mapuri, line 72: even for provisioning contiv services on a baremetal, we should be using a baremetal that is setup using `contiv/lab` and that I think falls in purview of this PR. 

I think we need to agree on a production environment usecase. And that usecase in my mind is that a user brings up a machine with a base os image (which is already provisioned using the `base` role) and then provisions contiv services on top. Do we agree?

PR #4, opened 2015-12-10
Fix etcd

erikh:
This apparently includes the ceph-install ansible as well, I'm not entirely sure why but if we could let this slide I would appreciate it, as it needs to be in here anyways.

This upgrades etcd and makes some changes to the init script to make it more reliable over a restart. I have validated that these fixes work, but the merge here iwth the ceph-install code means I'm revalidating while you guys are reviewing.

Thanks, if @jojimt and @mapuri have time to review that'd be really great.


roles/base/tasks/main.yml
```diff
@@ -43,13 +43,13 @@
     dest: /usr/local/
 - name: install Golang v1.5.1
   shell: chdir=/usr/local/ creates=/usr/local/go/bin/go tar xfvz go1.5.1.linux-amd64.tar.gz
-- name: download etcd v2.1.1
+- name: download etcd v2.2.2
   get_url: 
     validate_certs: no
-    url: https://github.com/coreos/etcd/releases/download/v2.1.1/etcd-v2.1.1-linux-amd64.tar.gz
+    url: https://github.com/coreos/etcd/releases/download/v2.2.2/etcd-v2.2.2-linux-amd64.tar.gz
```
mapuri, line 49: @erikh do we know if this version bump is earning us something. My concern is that it might have impact on volplugin and netplugin as they largely depend on it and we have been doing fine so far with v2.1. Is it something we need feedback from @shaleman or @jainvipin? Otherwise, I am ok with this version bump.

roles/etcd/templates/etcd.j2
```diff
@@ -64,21 +72,9 @@ start)
     ;;
 
 stop)
-    {% if run_as == "worker" -%}
-    echo "==> no 'stop' action for proxy"
-    {% else -%}
```
mapuri, line 69: I think we need to keep this logic for the scenarios where we incrementally add members to the cluster and for `worker` nodes. May be we should add a `{% elif etcd_init_cluster -%}` block like for the `service start` case above and take no action in that block?

roles/etcd/templates/etcd.j2
```diff
@@ -64,21 +72,9 @@ start)
...
-    if [ "$out" != "" ]; then
-        echo "==> removing member: " $out
-        etcdctl member remove $out
-    fi
-    {% endif -%}
     ;;
 
 post-stop)
-    #XXX: is there a case when we should not cleanup the data dir on stop?
-    rm -rf $ETCD_DATA_DIR
```
mapuri, line 81: Is this intentional? My understanding was that we need to remove the data-dir on service stop, so just curious if you were seeing any issue with data dir being removed.

mapuri:
@erikh I think I have a slightly different approach of handling member addition/deletion. It seems to be working for me. I think it is slightly more correct as we are able to remove the membership when etcd service is stopped as it gracefully removes the member from cluster. WDYT?

https://github.com/mapuri/ansible/commit/b37168f9710620522bdc956945a2e5b270861fe2

roles/base/tasks/main.yml
jojimt, line 49: Would be a good idea to test this out on netplugin before comming this bump.

roles/etcd/templates/etcd.j2
```diff
@@ -35,7 +35,15 @@ start)
...
-    export ETCD_INITIAL_CLUSTER_STATE=new
+
+    if [ ! -f /var/tmp/etcd.existing ]
+    then
+      export ETCD_INITIAL_CLUSTER_STATE=new
+      touch /var/tmp/etcd.existing
+    else
+      export ETCD_INITIAL_CLUSTER_STATE=existing
+    fi
+
```
jojimt, line 46: Would this impact an etcd upgrade? Let's see we are going from version 2.1 to 2.2. In that case, should we clean this up?

roles/base/tasks/main.yml
erikh, line 49: It's not committed to master yet.

roles/etcd/templates/etcd.j2
erikh, line 46: 2.1 to 2.2 is a seamless upgrade. 
https://github.com/coreos/etcd/releases/tag/v2.2.0

roles/base/tasks/main.yml
erikh, line 49: Oh I see what you mean, derp. Sorry. I don't think this is necessary but I'm using the 2.2 client so I figured it might help things. I can back it out.

roles/etcd/templates/etcd.j2
erikh, line 69: ok.

erikh:
@mapuri your solution does look better and I will try it out shortly.

erikh:
Yes I can confirm this patch works with my new tests and I'd prefer to 
use it. I will close this PR and allow you to make yours, and we can 
review it fast so it can be vendored.

Thanks @mapuri and @jojimt for reviewing.

mapuri:
thanks for verifying @erikh ! 

I have created #5 with the patch.

PR #5, opened 2015-12-10
fix etcd member addition for cluster init scenario

mapuri:
When etcd service is stopped we delete the node's membership from etcd cluster for a graceful exit. However, when the service is started again the member was not getting added for a node that has been added to the cluster as part of cluster init scenario.

This patch tries to alleviate that by doing a member add in above scenario.


erikh:
LGTM

erikh:
can we merge this? I'm depending on it.

mapuri:
yep, I think so. It might need to be improved if we find more cases. 

Will merge this.

PR #6, opened 2015-12-13
ansible: do not run the ceph bootstrap process on more than one host

erikh:
/cc @mapuri 

I'm fairly certain this fixes the issue with writing to the host more than once; let me know if you have any objections.

### DIFF
--- a/roles/ceph-mon/tasks/ceph_keys.yml
+++ b/roles/ceph-mon/tasks/ceph_keys.yml
@@ -4,6 +4,7 @@
 - name: wait for client.admin key exists
   wait_for:
     path: /etc/ceph/ceph.client.admin.keyring
+  run_once: true
 
 - name: create ceph rest api keyring
   command: ceph auth get-or-create client.restapi osd 'allow *' mon 'allow *' -o /etc/ceph/ceph.client.restapi.keyring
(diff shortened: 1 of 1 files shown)


roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -13,17 +14,20 @@
   when:
     cephx and
     groups[restapi_group_name] is defined
+  run_once: true
```
mapuri, line 17: @erikh, not sure if this is a big one but I am wondering if this will prevent the key from generated on all monitor hosts (except one). Do we copy this key later to the monitor hosts? Otherwise we may not be able to make REST API call on monitor hosts other than the first one.

roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -33,6 +37,7 @@
     group: root
   with_items:
     - "{{ ceph_keys.stdout_lines }}"
+  run_once: true
```
mapuri, line 40: Again I might be wrong but we may need to set the file permissions on all monitor hosts, right?

roles/ceph-mon/tasks/ceph_keys.yml
```diff
@@ -54,3 +60,4 @@
     group: root
     mode: 0755
   when: ansible_distribution_release == 'precise'
+  run_once: true
```
mapuri, line 63: this script seems good to have and might be ok to copy on all monitor hosts?

mapuri:
thanks @erikh . The changes look good. I do have some clarifications otherwise this should be good to merge.

roles/ceph-mon/tasks/ceph_keys.yml
erikh, line 17: Yeah, it's copied in the ceph-mon and ceph-osd roles.

mapuri, line 17: ah, ok... this should be ok then.

erikh:
whoops! deleted the branch, sorry, we'll have another pass in a minute

PR #7, opened 2015-12-17
Misc Fixes

mapuri:
- replaced ansible_hostname with inventory_hostname to allow running palybooks in environments where the hostname doesn't necessarily match the inventory name.
- made etcd master addr and name variables undefined to fail early than resulting in a mis-configured cluster.

@erikh PTAL. I think we have discussed about `inventory_hostname` name issue in past but wasn't sure if you already had something like this in works.

### DIFF
--- a/group_vars/all
+++ b/group_vars/all
@@ -7,8 +7,8 @@ cephx_require_signatures: false # Kernel RBD does NOT support signatures!
 cephx_cluster_require_signatures: true
 cephx_service_require_signatures: false
 
-node_name: "{{ ansible_hostname }}"
-node_addr: "{{ hostvars[ansible_hostname]['ansible_' + monitor_interface]['ipv4']['address'] }}"
+node_name: "{{ inventory_hostname }}"
+node_addr: "{{ hostvars[inventory_hostname]['ansible_' + monitor_interface]['ipv4']['address'] }}"
 
 crush_location: false
 osd_crush_location: "'root={{ ceph_crush_root }} rack={{ ceph_crush_rack }} host={{ ansible_hostname }}'"
--- a/roles/etcd/templates/etcd.j2
+++ b/roles/etcd/templates/etcd.j2
@@ -41,9 +41,9 @@ start)
         export ETCD_INITIAL_CLUSTER="
         {%- for host in groups[etcd_peers_group] -%}
         {%- if loop.last -%}
-        {{ hostvars[host]['ansible_hostname'] }}=http://{{ hostvars[host]['ansible_' + etcd_peer_interface]['ipv4']['address'] }}:{{ etcd_peer_port1 }},{{ hostvars[host]['ansible_hostname'] }}=http://{{ hostvars[host]['ansible_' + etcd_peer_interface]['ipv4']['address'] }}:{{ etcd_peer_port2 }}
+        {{ hostvars[host]['inventory_hostname'] }}=http://{{ hostvars[host]['ansible_' + etcd_peer_interface]['ipv4']['address'] }}:{{ etcd_peer_port1 }},{{ hostvars[host]['inventory_hostname'] }}=http://{{ hostvars[host]['ansible_' + etcd_peer_interface]['ipv4']['address'] }}:{{ etcd_peer_port2 }}
         {%- else -%}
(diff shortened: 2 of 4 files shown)


erikh:
man, we can't apply this patch and track upstream. I don't know what else to say other than convince the ceph-ansible maintainers it's a good idea, otherwise we're basically abandoning all ability to keep in sync with the requirements of the newest ceph versions, updates to the ansible itself, etc.

erikh:
I'm having to move our ceph-install stuff around due to comments in https://github.com/ceph/ceph-ansible/pull/463 and I don't think something this fundamental will swing with them, naturally. Perhaps we can find another way of accomplishing the same thing?

mapuri:
> Perhaps we can find another way of accomplishing the same thing?

I think that is going to be tough without a change in ceph playbooks.

I will create a PR upstream in ceph repo as well for this change and start a thread there to get the opinion of maintainers. I think that is a right way to go about this.

In the interim, do you think the wrapper playbooks that mentioned earlier can help us in some way? We do need an ability to be able to run in environments where hostnames don't always match inventory names.

erikh:
Works for me on the PR. Link it here please when you create it.

erikh:
Regarding wrapper playbooks, I don't know if that's going to do it. We're going to need to set this var somewhere, and it's depending on the hostname of the host, not a var we can manipulate.

One option that we could do is set the hostname on OUR hosts to what the inventory_hostname is set to, therefore syncing up the two. Would that be a workable solution?

mapuri:
> One option that we could do is set the hostname on OUR hosts to what the inventory_hostname is set to, therefore syncing up the two. Would that be a workable solution?

No, that doesn't work in a cluster environment with baremetals where we don't get to control the hostname.

erikh:
What's the limitation on setting the inventory hostname to the actual name of the box?

Also, when would you not be able to change a hostname on a bare metal machine, but not also already know it in advance? I don't understand.

mapuri:
> What's the limitation on setting the inventory hostname to the actual name of the box?

The main limitation is that hostnames can't be guaranteed to be unique in a cluster of machines. But inventory names on other hand are guaranteed to be unique (as they are there in inventory file identifying each host). One way to derive a unique inventory-name is to append server's serial number to it's hostname. This is is what I do in cluster to derive the inventory-name.

> when would you not be able to change a hostname on a bare metal machine, but not also already know it in advance? 

If my above argument is reasonable, then you would notice that in a cluster we get to manage the node after it has been discovered i.e. it's hostname has been set. So eventhough the inventory-name get's derived after the hostname is set, the cluster-manager itself doesn't participate in setting the hostname of the server.

mapuri:
@erikh this is ready for another look. Please take a look whenever you get a chance.

So after some digging ceph-playbook don't really need to change to use inventory-hostname. I am pasting out from slack conversation for context:

```
mapuri [4:50 PM] 
hey @erikh regarding https://github.com/contiv/ansible/pull/7#issuecomment-165573794
​[4:50] I diged in further and I think ceph -playbooks are fine
​[4:50] basically they use both inventoty-hostname and ansible-hostname
​[4:51] but they only use ansible-hostname while iterating over hostvars[] dictionary
​[4:51] which guarantees ansible_hostname to be available
​[4:51] so, I will update the patch and remove ceph related changes
erikh [4:57 PM] ok
```

mapuri:
@erikh can you please review this when you get a chance?

group_vars/all
```diff
@@ -7,11 +7,11 @@ cephx_require_signatures: false # Kernel RBD does NOT support signatures!
...
 cephx_service_require_signatures: false
 
-node_name: "{{ ansible_hostname }}"
-node_addr: "{{ hostvars[ansible_hostname]['ansible_' + monitor_interface]['ipv4']['address'] }}"
+node_name: "{{ inventory_hostname }}"
+node_addr: "{{ hostvars[inventory_hostname]['ansible_' + monitor_interface]['ipv4']['address'] }}"
 
 crush_location: false
-osd_crush_location: "'root={{ ceph_crush_root }} rack={{ ceph_crush_rack }} host={{ ansible_hostname }}'"
+osd_crush_location: "'root={{ ceph_crush_root }} rack={{ ceph_crush_rack }} host={{ inventory_hostname }}'"
```
erikh: I think this still needs to be `ansible_hostname` here.

erikh:
LGTM once that last thing is resolved.

group_vars/all
mapuri: good catch...looks like I missed this...let me fix this.

mapuri:
@erikh I took care of the comment. PTAL

erikh:
LGTM

PR #8, opened 2015-12-18
Netplugin fixes

erikh:
This incorporates several fixes required for netplugin to run under the `centos71-netplugin` vagrant box in Vagrant, under a multi-host scenario.

It also adds a small fix to install golang 1.5.2 only once per provision.


vvb:
@erikh while the changes LGTM, there are some other differences in how netplugin/etcd gets invoked here in the ansible repo.

contiv/ansible:
    /usr/bin/netplugin
    /usr/bin/etcd

contiv/netplugin:
    /opt/gopath/bin/netplugin -docker-plugin
    etcd --name netplugin-node1 --data-dir /tmp/etcd --listen-client-urls http://0.0.0.0:2379,http://0.0.0.0:4001 --advertise-client-urls http://192.168.2.10:2379,http://192.168.2.10:4001 --initial-advertise-peer-urls http://192.168.2.10:2380,http://192.168.2.10:7001 --listen-peer-urls http://192.168.2.10:2380 --initial-cluster netplugin-node1=http://192.168.2.10:2380,netplugin-node1=http://192.168.2.10:7001,netplugin-node2=http://192.168.2.11:2380,netplugin-node2=http://192.168.2.11:7001 --initial-cluster-state new

Are the args initialised somewhere else internally? I couldn't find it in the repo atleast

roles/contiv_network/tasks/main.yml
```diff
@@ -41,13 +29,20 @@
 - name: install contivctl
   shell: source /etc/profile.d/00golang.sh && go get github.com/contiv/contivctl
 
-- name: install docker-compose
-  shell: pip install docker-compose
+- yum: name=python-pip
```
mapuri: I think the task name went missing here. It's just a convention that we have been following to add task names, so may be we should just follow that here too.

Also slightly unrelated, and might not be in the purview of this PR. But I think we can move the docker-compose related installation tasks to the docker role. Would you be ok to do that as part of this PR?

roles/contiv_network/tasks/main.yml
```diff
@@ -41,13 +29,20 @@
...
-- name: install contiv-compose
-  shell: >
-     source /etc/profile.d/00golang.sh && \
-     go get github.com/contiv/libcompose && \
-     cd $GOPATH/src/github.com/contiv/libcompose && \
-     make binary && \
-     ln -s `pwd`/bundles/libcompose-cli_linux-amd64 $GOPATH/bin/contiv-compose
+- pip: name=docker-compose state=latest extra_args=--upgrade
+
+- name: copy environment file for netmaster
```
mapuri, line 28: this is a minor nit. I think as part of edits to this file the block for netmaster configuration got moved to the end of file starting from this line onwards. We should move it back to it's original position because the block that installs compose and other stuff above has a comment that things there need to be moved out of this file. So a reader in future might get an idea that even netmaster needs to be moved out of this file which can cause confusion :)

site.yml
```diff
@@ -18,11 +18,14 @@
   environment: env
   roles: 
   - { role: base }
+  - { role: vagrant }
```
mapuri, line 21: `vagrant` role is a pretty nice introduction :+1: 

I can see that we might be able to use it for baremetal environments as well that run behind proxy and for setting up our devtest machines. In future, I am thinking we could enhance this role to be a little more generic for setting up bashrc for any user. Having this role gives me that way! Hope you will be ok with me making enhancements to this role.

mapuri:
@erikh LGTM other than a few minor comments

@vijayvikrant 

> Are the args initialised somewhere else internally? I couldn't find it in the repo atleast

the etcd args are set by the etcd role, checkout the `https://github.com/contiv/ansible/blob/master/roles/etcd/templates/etcd.j2` template file that sets up etcd in different scenarios viz. cluster-init, member-add and proxy. Specific to Erik's change for `volplugin-test` host-group cluster-init scenario will get executed. Let me know if you will need help navigating through it.

roles/contiv_network/tasks/main.yml
erikh, line 28: it can't be started before netplugin, which was the reasoning for moving it. I will move the comment(s).

site.yml
erikh, line 21: Of course :)

roles/contiv_network/tasks/main.yml
erikh: yeah, sure.

erikh:
This is ready for another pass, PTAL

mapuri:
LGTM

PR #10, opened 2015-12-21
add stop script for collins service

mapuri:
@erikh PTAL when you get a chance. This is needed to be able to restart collins as part of cluster manager systemtests but might be helpful otherwise as well.


erikh:
LGTM

Issue #11, opened 2015-12-24
Proposal: reorganize/cleanup the base role


mapuri:
**Background:**
At some point when we decided to use ansible for provisioning all our projects we started off with the ansible playbooks used by our packer builds. At that point `base` role was added to download and install multiple binaries. It was pretty much everything that packer builds were installing like baked in go, docker, etcd, ovs and so on.

**Fast Forward to requirements:**
Overtime as we have gained some more experience adding ansible plays for our own services, the existing organization of `base` role has gotten a bit unaligned with the general organization of other plays like:
1. not all service specific dependencies are always installed together. Just to state one instance, docker binary get's installed in base but the service is setup as part of `docker` role; same is true with etcd and ovs.
2. also overtime we have realized there are two broad environments that we are going to have a need to support with our ansible scripts for deploying Contiv services viz. **production** and **development**. A **production** environment would be a **stock** Ubuntu15.04 or Cento7.1 or in future a RHEL OS image loaded by the customer and then provisioned using our ansible playbooks according to their deployment. A **development** environment is similar to customer environment but we want some packages to be pre-installed in order to save time when bringing up the development environment. An example of **development** environment will be a vagrant box that we use for testing in our projects or the host machines that we use for development.

Resolving the above will hopefully help us get more clarity on how we can use our playbooks in customer/demo environments and development environments.

**Proposal:**

To achieve the above two requirements, I propose the following:
- move the service specific dependencies to respective roles that take care of provisioning the service. This shall address **1.** above.
- introduce a new role called `dev` that shall take care of pre-installing the requirements needed for the **development** environment but it shall reuse the tasks from underlying services. This shall address **2.** above.

Also see the associated patch of something in works in my fork here: https://github.com/mapuri/ansible/tree/devrole

This shall give a more concrete idea of what I have in mind. I will create a formal PR once I have it more polished and tested. Feel free to take a look and suggest other cleanups as well that you might have had in mind.

cc @jainvipin @erikh

mapuri:
closed by #14

PR #12, opened 2015-12-25
Netplugin ansible changes

dvavili:
Create new hostgroup netplugin-node
- Pulls skydns container as part of hostgroup
- Installs released version of netmaster/netplugin when deply_released_version: true
  Fix minor bugs with source and installing docker-compose


roles/contiv_network/tasks/main.yml
```diff
@@ -1,44 +1,10 @@
...
-- name: start netmaster
-  shell: systemctl daemon-reload && systemctl start netmaster
-  when: run_as == "master"
-  run_once: true
-
-# FIXME: need to move the following to correct roles -- also needs release tooling
-- name: install contivctl
-  shell: source /etc/profile.d/00golang.sh && go get github.com/contiv/contivctl
+- name: pull dns container image
+  docker:
```
mapuri: do we really need to use `docker` module? we can do this using shell, right?

roles/contiv_network/tasks/main.yml
```diff
@@ -1,44 +1,10 @@
...
+    image: skynetservices/skydns
+    pull: always
+    state: absent
+
+- include: deploy_release.yml
+  when: deploy_released_version is defined and deploy_released_version == true
```
mapuri: as we were discussing yesterday this logic is better done in Vagrant environment itself. By default ansible shall always deploy released version and we shall not need a variable like `deploy_released_version`. Do this distinction in ansible kind of beats the purpose of having ansible to deploy the service.

And with that we won't need a separate `deploy_release.yml` as well.

roles/docker/tasks/main.yml
```diff
@@ -24,5 +24,15 @@
   shell: systemctl daemon-reload && systemctl stop docker && systemctl start docker-tcp.socket && systemctl start docker
   when: docker_env_conf_result.changed or docker_tcp_conf_result.changed or docker_svc_conf_result.changed
 
+#NOTE: Without this step, pip crashes after installing docker-py/docker-compose
+#      Move to appropriate role later
+- name: upgrade pip
```
mapuri: both these steps can go away if we don't use `docker` module as we discussed yesterday.

mapuri:
@DivyaVavili thanks for the changes. 

Look pretty good. I have a few minor nits hopefully we can resolve and then merge.

roles/contiv_network/tasks/main.yml
jainvipin: If all use cases for netplugin reuquire dns container, it is perhaps better to pull it in ansible.
Starting can be done by shell, because of the ansible variables it might need. Arguably if no runtime params are needed to be passed, we can run it as part of ansible too. That is an infra service that must be started alongside rest of the things (like netplugin), etc.

roles/contiv_network/tasks/main.yml
jainvipin: @mapuri - if we do pull in the shell calling ansible, we'd have to replicate this in all shells; ideally shell calling ansible should be just left with some customization that is needed to call ansible itself, if any. WDYT?

jainvipin:
where are we starting the dns container? Is that not part of ansible?
Are vagrant changes supposed to be outside this PR?

roles/contiv_network/tasks/main.yml
dvavili: @mapuri : So, I did look into this and the reason I prefer using docker module is that prevents it from downloading again if it's already present... I ran a couple of tests to confirm this... This would save some time and also prevent issues if the proxy server is not reachable and the like

mapuri: @jainvipin I was referring to use `shell` module instead, that simplifies things imo like we don't need to pull extra dependency like docker-py etc

roles/docker/tasks/main.yml
dvavili: Upgrading pip is necessary here as docker-compose module which is done after this task is also crashing pip... the issue is not visible immediately, but when pip module is used later, it shows up. It might be difficult to debug this issue later without much context.

roles/contiv_network/tasks/main.yml
mapuri: Also in future, we might want to run the `sky-dns` container just like a service... in that case `docker` module might actually not be good as we use systemd units to start our sevices to get consistent logging and monitoring experience. For instance, this is how ended up doing it for `collins` container needed for cluster: https://github.com/contiv/ansible/blob/master/roles/contiv_cluster/tasks/main.yml#L4

dvavili:
@jainvipin Starting the container is taken care of by tenant creation in netmaster. We just need to pull the skydns container for netmaster to do the necessary actions from then on

roles/contiv_network/tasks/main.yml
dvavili: @mapuri ... From what I understand, the next set of changes for ACI also requires the docker module to handle aci-gw container... So, even if I were to remove it now, it's going to be used soon. And as you mentioned, when we move skydns (and aci-gw, if required) as services, this can be removed. But I guess we'll need it till then...

site.yml
```diff
@@ -61,5 +61,15 @@
   - { role: etcd, run_as: worker }
   - { role: ceph-osd, mon_group_name: service-master, osd_group_name: service-worker }
   - { role: swarm, run_as: worker }
-  - { role: contiv_network, run_as: worker }
+  - { role: contiv_network, run_as: worker, deploy_released_version: true }
   - { role: contiv_storage, run_as: worker }
+
+# netplugin-node hosts set up netmast/netplugin in a cluster
+- hosts: netplugin-node
```
jainvipin, line 68: @DivyaVavili - netplugin-mode is run as master, how about non master nodes?

roles/contiv_network/tasks/main.yml
dvavili: Also, the only other change that is added on top of this is to upgrade pip and as mentioned in the below comment, it is needed for docker-compose as well...

mapuri: @DivyaVavili 

> the reason I prefer using docker module is that prevents it from downloading again if it's already present

`docker pull` command will also not pull if the image already exists, so it should be the same as docker module. 

>  also prevent issues if the proxy server is not reachable and the like

I am not sure if this is a big gain, I would still prefer simplicity over this, especially since we are going to burn in skydns images in our dev environments using this ansible.

Also see my comment on starting sky-dns as a separate service unit in final deployment. We won't gain much using docker module as this pull task is only for a short term.

roles/contiv_network/tasks/main.yml
dvavili: The netplugin-node hostgroup deploys the released version by default, and Vagrantfile handles it so that it is not using the released version. I can do the reverse if needed.

jainvipin:
LGTM, will try this out

roles/contiv_network/tasks/main.yml
mapuri: yeah, the reverse is better because vagrant is the only environment where we will want to mount developer's workspace in the vm while ansible will be mostly used for deploying the service from a released tree.

roles/contiv_network/tasks/main.yml
mapuri: > From what I understand, the next set of changes for ACI also requires the docker module to handle aci-gw container... 

Let's take that when it comes. From what I remember, I think aci container is an immediate candidate to be started as a systemd unit since it is not started internally by netmaster.

dvavili: So, this is the logs from the second run of provisioning for the nodes. Shell module seems to be repulling the image even if it exists. Also, this might be important, because a lot of the complaints that I have been hearing is because of re-pulling this image during vagrant bringup:
When using docker module
TASK: [contiv_network | pull dns container image] *****************************
ok: [netplugin-node1]
ok: [netplugin-node2]

When using shell module:
TASK: [contiv_network | pull dns container image] *****************************
changed: [netplugin-node2]
changed: [netplugin-node1]

The second reason is that like I mentioned, I will definitely have to reintroduce this docker module within a week when aci ansible changes come in. We will definitely remove this when we move it to a service model. Till then I prefer keeping this.

roles/docker/tasks/main.yml
mapuri: I think `base` already installs `pip` so we should be ok there. And I don't think we need docker-py as such for compose.

I just tried installing this host-group with this change reverted. The docker-compose installation worked for me after `base` is run on the node, so we should be ok there.

site.yml
dvavili, line 68: @jainvipin: netmaster is run only once in a cluster by using run_once option.

roles/contiv_network/tasks/main.yml
mapuri: > this is the logs from the second run of provisioning for the nodes. Shell module seems to be repulling the image even if it exists.

`changed` status for shell module need not mean it downloaded the image again. It will only indicate that the module was run. May be measuring the time to run the task itself will be a good measure. Let me see if I can instrument it to prove to myself.

> The second reason is that like I mentioned, I will definitely have to reintroduce this docker module within a week when aci ansible changes come in.

I am not sure if that decision is made yet. And I will make sure that it goes through a review before that decision is made :).

dvavili: Sure, your comments are welcome even then :) 
But I don't want to be chasing proxy issues after this change goes in (which has been going on for a month now) and I'm comfortable with the docker module than the shell module for handling this issue.
If you still feel shell is better, I can make a shift.

mapuri: yes, shell is simple and better for this.

I just tried it out. It won't pull the image again as I saw in my quick experiment below.

```
TASK: [contiv_network | pull dns container image] *****************************
changed: [cluster-node1]

TASK: [contiv_network | debug msg="{{ pull.stdout }}"] ************************
ok: [cluster-node1] => {
    "msg": "Using default tag: latest\nlatest: Pulling from skynetservices/skydns\nDigest: sha256:6f8a9cff0b946574bb59804016d3aacebc637581bace452db6a7515fa2df79ee\nStatus: Image is up to date for skynetservices/skydns:latest"
}
```

The solution for proxy issue is to pre-bake the container in the vagrant box. And that should not be solved in ansible imo. I am working on something that might help us get development images: https://github.com/contiv/ansible/issues/11

roles/docker/tasks/main.yml
erikh: This is actually pretty important for docker-compose use, but I have a 
patch for it that needs to be submitted to contiv/ansible still. if we 
end up removing this patch d-c may not work without it.

erikh:
@DivyaVavili is this being tested against hte 0.3.x boxes or the 0.5.x boxes?

dvavili:
@erikh: These are being tested against 0.3.x boxes for now

dvavili:
@jainvipin @mapuri @erikh: Please review the latest changes.
To summarize:
- Removed docker module for DNS pull and pulling the container using shell module
- Included contiv_dev(false by default) and contiv_demo(true by default) variables in contiv_network to handle scenarios like demo/dev/test environment. 
- By default, netplugin-node hostgroup will launch released binaries
- In a dev environment, contiv_dev is set to true and it picks up binaries from local repositories and launches them as services
- In a test environment; where our test scripts launch/cleanup netmaster/netplugin; contiv_demo is set to false and prevents launching of the binaries automatically

erikh:
@DivyaVavili you need to test these on 0.5.0 before they can be merged.

erikh:
Please rename `contiv_dev` and `contiv_demo` to `netplugin_dev` and `netplugin_demo`.

jainvipin:
@erikh - since the variables are acted upon within contiv_network, I am okay with `contiv_dev` but perhaps `contiv_demo` could be changed to `contiv_user` - this could allow us to keep one variable for development and another one for user. What if someone really wants to work on both, or jointly? A bit of stretch but quite conceivable that minor changes may be needed in one vs. another.

@DivyaVavili - can we avoid installing golang, etc. when in `contiv_user` mode? Or is task belonging to the base?

@DivyaVavili - I also assume that picking up the latest skydns images would be all okay... i.e. no specific version to pick up there.

LGTM otherwise...

erikh:
@jainvipin they are not scoped to `contiv_network`, they are global and will be used like globals. If we don't do this, we should find some way to make sure this fits with volplugin somehow.

roles/contiv_network/tasks/main.yml
```diff
@@ -1,44 +1,13 @@
...
-- name: start netmaster
-  shell: systemctl daemon-reload && systemctl start netmaster
-  when: run_as == "master"
-  run_once: true
-
-# FIXME: need to move the following to correct roles -- also needs release tooling
-- name: install contivctl
-  shell: source /etc/profile.d/00golang.sh && go get github.com/contiv/contivctl
+- include: deploy_binaries.yml
+  when: contiv_demo
```
mapuri: @DivyaVavili  as we have discussed in slack previously I don't think we should make make deployment conditional in ansible. It should be done in the project as deploying a dev environment should be limited to the project.

Let's chat on slack, I can help to figure that out.

roles/docker/tasks/main.yml
```diff
@@ -24,5 +24,12 @@
   shell: systemctl daemon-reload && systemctl stop docker && systemctl start docker-tcp.socket && systemctl start docker
   when: docker_env_conf_result.changed or docker_tcp_conf_result.changed or docker_svc_conf_result.changed
 
+#NOTE: Without this step, pip crashes after installing docker-compose
+#      Move to appropriate role later
+- name: upgrade pip
```
mapuri: the `base` role already installs pip. This task should not be needed if `base` is run before running `docker` role.

roles/contiv_network/tasks/main.yml
```diff
@@ -1,44 +1,13 @@
...
+  ignore_errors: yes
 
-- name: download netmaster and netplugin
-  get_url:
-    validate_certs: no
-    url: https://github.com/contiv/netplugin/releases/download/v0.0.0-12-11-2015.20-54-40.UTC/netplugin-v0.0.0-12-11-2015.20-54-40.UTC.tar.bz2
-    dest: /tmp/contivnet.tar.bz2
+- name: pull dns container image
+  shell: docker pull skynetservices/skydns
+  when: "'No such image' in docker_inspect_result.stderr"
```
mapuri, line 10: the pull should be a noop if the image exists, we don't need this conditional, right?

roles/docker/tasks/main.yml
erikh: the `--upgrade` `docker-compose` needs will break pip. We should probably move this to the base role, though.

roles/contiv_network/tasks/main.yml
erikh, line 10: On Thu, Dec 31, 2015 at 10:43:17AM -0800, Madhav Puri wrote:

> > -- name: download netmaster and netplugin
> > -  get_url:
> > -    validate_certs: no
> > -    url: https://github.com/contiv/netplugin/releases/download/v0.0.0-12-11-2015.20-54-40.UTC/netplugin-v0.0.0-12-11-2015.20-54-40.UTC.tar.bz2
> > -    dest: /tmp/contivnet.tar.bz2
> >   +- name: pull dns container image
> > -  shell: docker pull skynetservices/skydns
> > -  when: "'No such image' in docker_inspect_result.stderr"
> 
> the pull should be a noop if the image exists, we don't need this conditional, right?

It's a significant, unnecessary network hit and would save a little time for a
little code, so I am +1 for this part of the patch.

-Erik

roles/docker/tasks/main.yml
mapuri: hmm, somehow for me docker-compose installs and works fine if I run `base`. 

Not sure what am I missing, but I am ok installing this as long as we see the need for it.

roles/contiv_network/tasks/main.yml
mapuri, line 10: sure, sgtm. I was hoping we could depend on docker daemon to optimize that for us i.e. not pull an image if it already exists on the machine. I tried looking around docker code and it does appears that it tries to reach out to registry for every pull, so may be this optimization is good.

dvavili:
@erikh: I've tested it out with 0.5.0 boxes and it looks good. Also made changes to change vars from contiv_dev(/demo) to netplugin_dev(/demo)..

dvavili:
@jainvipin: Skydns pull will pick up the latest and not tied to a specific version.
With regards to not setting up golang and other environment, I see a new role dev being introduced by @mapuri ... We could work something off that, but netplugin_node setup might not be the right place to handle it...

jainvipin:
@DivyaVavili 

> @jainvipin: Skydns pull will pick up the latest and not tied to a specific version.

This means you are signing up to say that it will work with latest - always; which may be okay for now until we discover something breaking

> With regards to not setting up golang and other environment, I see a new role dev being introduced by

sure, let's discuss this as a part of @mapuri's checkin

mapuri:
@DivyaVavili thanks for taking care of comments and your patience with this. Much appreciated!!

LGTM

erikh:
I'll try to give this a final pass later today.

erikh:
ok. Now that I've made you do all this, please test again on `contiv/centos72` version `0.1.1`. :) Not necessary for merge, but please do test as we'll be heading this way with our redhat boxes.

roles/contiv_network/tasks/main.yml
```diff
@@ -41,4 +49,4 @@
 
 # FIXME: need to move the following to correct roles -- also needs release tooling
 - name: install contivctl
-  shell: source /etc/profile.d/00golang.sh && go get github.com/contiv/contivctl
+  shell: . /etc/profile.d/00golang.sh && go get github.com/contiv/contivctl
```
erikh, line 52: is this change really necessary?

erikh:
LGTM other than my comment. Since `.` and `source` are equivalent, let's not add this, I don't believe it's listed in the commits anyhow.

dvavili:
@erikh: Ran into issues when using source command... While it works on a centos box, it fails for ubuntu box. I think the reason is that on Ubuntu, dash shell is used by default (on which the . notation is the norm)... Changing it helped. @mapuri is aware of this issue...

mapuri:
yes, what we noticed was that `source` seems to be not available in '/bin/sh' shell on Ubuntu (but it is there on Centos) that the shell module seems to use.

**\* On Ubuntu ***

```
ladmin@swvm168:~/goroot/src/github.com/contiv/cluster$ /bin/sh
$ source
/bin/sh: 1: source: not found
$ exit
```

**\* On Centos ***

```
[vagrant@cluster-node1 ~]$ /bin/sh
sh-4.2$ source
sh: source: filename argument required
source: usage: source filename [arguments]
sh-4.2$
```

Alternatively we can set the `executable` attribute of the shell module to `/bin/bash` but since `.` works on all shells that shall be fine too.

erikh:
LGTM

PR #13, opened 2015-12-28
base/tasks/main.yml: don't download Go if present

unclejack:


erikh:
LGTM

erikh:
@mapuri PTAL

roles/base/tasks/main.yml
```diff
@@ -38,13 +38,18 @@
 - name: install and start ntp
   shell: systemctl enable ntpd
   when: ansible_os_family == "RedHat"
+- name: check if golang is installed
+  stat: path=/usr/local/go/bin/go
+  register: golang_bin
 - name: download Golang v1.5.2
   get_url:
```
mapuri, line 45: I think we can use the `force=no` attribute of the `get_url` module to prevent a download if the file already exists.

roles/base/tasks/main.yml
```diff
@@ -38,13 +38,18 @@
...
     validate_certs: no
     url: https://storage.googleapis.com/golang/go1.5.2.linux-amd64.tar.gz
     dest: /tmp/golang.tar.gz
-- name: install Golang v1.5.1
+  when: golang_bin.stat.exists == false
+- name: install Golang v1.5.2
   shell: chdir=/usr/local/ creates=/usr/local/go/bin/go tar xfvz /tmp/golang.tar.gz
+  when: golang_bin.stat.exists == false
```
mapuri, line 52: I think the `creates=/usr/local/go/bin/go` attribute of the `shell` module will take care of not running this task if the binary already exists.

mapuri:
@unclejack, @erikh I think we can avoid the need for `stat` module if we can depend on the `force=no` and `creates` attributes of the `get_url` and `shell` modules respectively. WDYT?

roles/base/tasks/main.yml
unclejack, line 52: @mapuri Yes, it was doing that before, but I've decided to migrate it to the same stat check to make it easier to keep track of it.

roles/base/tasks/main.yml
unclejack, line 45: @mapuri This file is downloaded in /tmp. It won't exist when we provision the VMs and we'll download it if we don't use stat. I've already looked into the other options.

mapuri, line 45: ah, ok, makes sense, thanks for clarifying.

roles/base/tasks/main.yml
mapuri, line 52: ok, sgtm. May be we should also remove the `creates=/usr/local/go/bin/go` statement here as it is redundant with the `when` conditional.

unclejack:
@mapuri PTAL, `creates=` is gone now.

mapuri:
LGTM. thanks @unclejack

PR #14, opened 2015-12-31
reorganize the base role

mapuri:
- move package installations close to respective roles
- add dev role to make the development environment related pre-packaging more explicit. 'dev' role shall be used for putting in dev environment related dependencies like build tools or pre-packaging large binaries to save download times.

This addresses #11. This is more of organizational change, so I tried to minimize any major logic changes/fixes. I tried this out in cluster repo and looks fine.

cc/ @erikh @jainvipin 

Next step, we can continue to identify more tasks/roles to be includes as dependecies for `dev` in order to pre-package more things (like container images used by our projects etc) which will hopefully expedite bringing up our dev-test environments, while keeping our ansible roles production ready.

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -1,6 +1,10 @@
 ---
 # This role contains tasks for configuring and starting docker service
 
+# XXX: figure a way to download a specific docker version
+- name: install docker
+  shell: creates=/usr/bin/docker curl https://get.docker.com | bash
+
 - name: create docker daemon's config directory
   file: path=/etc/systemd/system/docker.service.d state=directory
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+#ide files
+*.swp
+
+# ansible fetch dir
+fetch/*
--- a/roles/base/tasks/main.yml
+++ b/roles/base/tasks/main.yml
@@ -13,9 +13,6 @@
     name: '*'
     state: latest
   when: ansible_os_family == "RedHat"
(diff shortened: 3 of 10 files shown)


mapuri:
ping @erikh @jainvipin

I would like to merge this if the proposed reorganization of roles looks good to you. PTAL when you get a chance.

roles/docker/tasks/main.yml
```diff
@@ -1,6 +1,10 @@
 ---
 # This role contains tasks for configuring and starting docker service
 
+# XXX: figure a way to download a specific docker version
```
erikh, line 4: http://releases.dockerproject.org/

erikh, line 4: The problem with this is that the package is what you want; without it you'll have to install the systemd units, etc by hand.

erikh:
LGTM. Strongly consider the comments, btu they are not necessary before merge.

roles/docker/tasks/main.yml
mapuri, line 4: thanks for the pointer @erikh let me look at the changes required to use a docker release.

Since we already modify the systemd units for docker it might not be that complex to download the release and set it up ourselves.

erikh:
Keep in mind there's some other packages, like iptables and bridging modules that must be installed. There are a few other things I'm sure these days.

mapuri:
ok, yeah briefly looking at this I think this will need a separate PR. Let me create a issue to track this and we can merge this first.

mapuri:
filed #16 to track docker role change, will merge this one now.

PR #15, opened 2016-01-01
Vmware fixes

erikh:
This helps our ansible cope with alternative centos environments: e.g., vmware. :)

@mapuri @DivyaVavili (because this covers a lot of your patch as well) and @jainvipin PTAL

### DIFF
--- a/roles/base/tasks/main.yml
+++ b/roles/base/tasks/main.yml
@@ -1,11 +1,24 @@
 ---
 # This role contains tasks for install base packages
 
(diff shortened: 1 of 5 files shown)


mapuri:
LGTM

erikh:
I'm going to merge this. I realize that it will conflict with both @mapuri's work and @DivyaVavili's. I am sorry to be so bullish here but given I'm going to be on vacation in less than a week for a month, I would like to get this in before bit rot sets in.

PR #23, opened 2016-01-08
Misc fixes and enhancements

### LINKED ISSUE #22: aci_demo_installer will fail if bzip2 is not installed in the worker nodes (Centos7)
Workaround: install bzip2 manually in the worker nodes

TASK: [contiv_network | install netmaster and netplugin] **********************
failed: [node2] => {"changed": true, "cmd": "tar vxjf /tmp/contivnet.tar.bz2", "delta": "0:00:00.007268", "end": "2016-01-07 08:08:07.476339", "rc": 2, "start": "2016-01-07 08:08:07.469071", "warnings": ["Consider using unarchive module rather than running tar"]}
stderr: tar (child): bzip2: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now

FATAL: all hosts have already failed -- aborting

PLAY RECAP ********************************************************************
           to retry, use: --limit @/root/site.retry

node2                      : ok=43  changed=24  unreachable=0    failed=1

mapuri:
- install ansible, packer, vagrant and virtualbox in respective roles
  - ansible is a base package as it will be used for cluster provisioning
  - packer, vagrant and virtualbox are dev environment related tools and will enable us to setup or baremetal host environments.
- fixed openvswitch installation for redhat
  - a task got leftover in `base` role as part of `dev` role change, which causes base role to fail.
- fixed yum package installation by separating out epel-release into it's own task
  - otherwise the packages that depend on it fail
- fixed apt package update task to just update the list and no try to update all installed packages
  - otherwise the task fails with conflicts
- install bzip2 in base packages (fixes #22)

/cc @erikh

### DIFF
--- a/roles/base/tasks/main.yml
+++ b/roles/base/tasks/main.yml
@@ -48,7 +48,6 @@
     - curl
     - git
     - mercurial
-    - openvswitch
     - gcc
     - perl
     - librbd1-devel
--- a/roles/base/tasks/main.yml
+++ b/roles/base/tasks/main.yml
@@ -4,7 +4,6 @@
 - name: upgrade system (debian)
   apt:
     update_cache: true
-    name: '*'
     state: latest
   when: ansible_os_family == "Debian"
 
--- a/roles/base/tasks/main.yml
+++ b/roles/base/tasks/main.yml
@@ -19,6 +19,7 @@
     name: "{{ item }}"
   with_items:
     - unzip
+    - bzip2
     - vim-nox
     - curl
     - python-software-properties
@@ -43,6 +44,7 @@
   with_items:
     - ntp
     - unzip
+    - bzip2
     - vim
     - curl
     - git
--- a/roles/base/tasks/main.yml
+++ b/roles/base/tasks/main.yml
@@ -7,12 +7,14 @@
     name: '*'
     state: latest
   when: ansible_os_family == "Debian"
+
 - name: upgrade system (redhat)
   yum:
     update_cache: true
     name: '*'
     state: latest
   when: ansible_os_family == "RedHat"
+
 - name: install base packages (debian)
   apt:
     name: "{{ item }}"
@@ -29,11 +31,17 @@
     - lshw
     - python-pip
   when: ansible_os_family == "Debian"
+
+# install epel-release first to ensure the extra packages can be installed later
+- name: install epel release package (redhat)
+  yum:
+    name: epel-release
+  when: ansible_os_family == "RedHat"
+
 - name: install base packages (redhat)
   yum:
     name: "{{ item }}"
   with_items:
-    - epel-release
     - ntp
     - unzip
     - vim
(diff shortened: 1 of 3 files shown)


roles/base/tasks/main.yml
```diff
@@ -4,20 +4,22 @@
...
     name: '*'
     state: latest
   when: ansible_os_family == "RedHat"
+
 - name: install base packages (debian)
   apt:
     name: "{{ item }}"
   with_items:
     - unzip
+    - bzip2
```
erikh, line 22: does this need to be installed on redhat too?

roles/base/tasks/main.yml
```diff
@@ -29,18 +31,24 @@
...
+  when: ansible_os_family == "RedHat"
+
 - name: install base packages (redhat)
   yum:
     name: "{{ item }}"
   with_items:
-    - epel-release
     - ntp
     - unzip
+    - bzip2
```
erikh, line 47: oh nevermind. :P

erikh:
I think we should start splitting up our work with includes and OS detection. e.g.:

``` yaml
- include: my_redhat_thing.yaml
- include: debian_thing.yaml
  when: ansible_os_family == "Debian"
```

I know it's a lot of work, but I'd really like this everywhere we control at least.

erikh:
I think the upgrade situation in the dev role... Might be painting ourselves into a corner if we aren't careful, as most of the packages here won't update if the files already exists on the system.

mapuri:
@erikh 

> I think we should start splitting up our work with includes and OS detection

good idea! I will create a issue and take that up next. It will be a good amount of work so better to keep in separate PR.

> I think the upgrade situation in the dev role

Can you explain a bit more? Do you mean that if I bump up a version of a binary in the ansible task and rerun the ansible, it will not upgrade for some? If so can you point me to an example may be I can think of a way.

erikh:
WFM on the new issue. I think packer in particular will be problematic, but we should be comparing versions of tools instead of existence of binaries.

mapuri:
> we should be comparing versions of tools instead of existence of binaries.

yes, that is a valid point for installing specific versions. I think with apt/yum packages it is easier to do that in same task, but with binaries that we download (like packer) and manually install we end requiring more task to stat and then get version. I think we can write a custom module or something to address this pattern. For now, I will add a separate task to check for packer version than checking the binary.

jainvipin:
@mapuri, @erikh 

> > we should be comparing versions of tools instead of existence of binaries.
> 
> yes, that is a valid point for installing specific versions. I think with apt/yum packages it is easier to do that in same task, but with binaries that we download (like packer) and manually install we end requiring more task to stat and then get version. I think we can write a custom module or something to address this pattern. For now, I will add a separate task to check for packer version than checking the binary.

If we don't compare the versions it is a non-starter IMO because if someone uses ansible to update to latest, it will never work. This happened to someone using our ansible while docker was already installed on one of the servers, however the version wasn't correct and thus created an issue.

ping @jojimt @DivyaVavili - as they ran into similar issue where upgrade didn't work

mapuri:
@jainvipin 

I agree. I am fixing packer (introduced by this PR) to do that. If folks have found other cases, I would suggest to file issues. I know there are other binaries that are installed similarly. I will also be reviewing the playbooks and fix them in subsequent cleanups.

jainvipin:
@mapuri - yes the plan was to submit a PR along with the changes needed for that instead of filing an issue; if you are working on fixing this for packer, then @DivyaVavili can take care of other binaries which she was planning on doing as a part of her subsequent PR.

mapuri:
@jainvipin, yes sgtm

@erikh this should be ready for another look, I have added version check for packer

roles/dev/tasks/main.yml
```diff
@@ -11,12 +11,90 @@
...
+  register: packer_version
+  ignore_errors: yes
+
+- name: download packer
+  get_url:
+    validate_certs: no
+    url: https://releases.hashicorp.com/packer/0.8.6/packer_0.8.6_linux_amd64.zip
+    dest: /tmp/packer_0.8.6_linux_amd64.zip
+    force: no
+  when: packer_version.stdout != "0.8.6"
```
erikh, line 35: does ansible have any strcmp() alike for this? If someone runs this ansible and has packer 0.9 installed, this may have dire consequences.

mapuri, line 35: By dire you mean that we will still go ahead and install 0.8.6, right? I think that is rather desirable as we would like to to make sure that the packages are actually at same version that we have validated here and support.

mapuri, line 35: just to close, based on slack conversation we agreed that current check should be good to ensure we install a validated version of tool.

PR #24, opened 2016-01-08
add prebake-for-dev tag for pre-baking binaries in dev images

mapuri:
This PR brings in:
- `prebake-for-dev` tag which shall be used to tag tasks for installing binaries from various roles in the `dev` role. This shall help limit dependencies for different contiv services to them while giving us a way to pre-bake certain binaries and save on dev/test bringup times. This tag shall be used by `packer` builds to pre-install binaries.
- added a Vagrantfile to test the `devtest` host-group which is only exercised as part of lab and build (unlike other host-groups)

Also with this change I am able to add a sanity job for our ansible repo (look at the Vagrantfile change).

/cc @erikh


mapuri:
@erikh this should be ready for a look now.

erikh:
@mapuri any way we can use includes here that key off a tag? We're doing it in a lot of places, I wonder if it'll be cleaner if we do it that way.

mapuri:
@erikh, when a tag is applied with `include` or the `role` clauses it get's applied to all the tasks in that role including it's dependencies. (http://docs.ansible.com/ansible/playbooks_tags.html#tags)

Right now, except for the `base` and `dev` roles all other roles just require a couple or so of their tasks to be tagged. ANd since `dev` includes these other roles as dependencies, we can't tag the entire `dev` role.

That leaves us just with `base` role which we can potentially tag just once at `devtest` host-group level. I had started off like that but then I moved to tagging individual tags for `base` thinking it to be consistent.

mapuri:
@erikh I have moved the tag for `base` role back up to the `devtest` hostgroup.

erikh:
I am wondering if this makes more sense:

``` yaml
- include: dev_tags.yaml
  tags: my_tag
```

mapuri:
@erikh yes, that organization makes it cleaner. I have reorganized `base` and `dev` tasks into os specific task files which are then included and tagged. PTAL when you get a chance. 

I think this will address #25 as well :)

erikh:
LGTM but I'd love to figure out a way to do this for the other roles as well, e.g. etcd.

mapuri:
thanks @erikh.

The other roles have a very selective few tasks in them that need to be tagged. We can have a file based approach there but it might be an overkill.

PR #31, opened 2016-01-13
os_agnostic_tasks.yml: bump Go to 1.5.3

unclejack:


mapuri:
LGTM

sanity seems to be failing during docker install. It is failing for me locally as well, will wait for a bit to see if docker remote repo recovers, otherwise I think this is good to merge. I am assuming volplugin and netplugin would be fine with this version bump.

mapuri:
@unclejack docker repo seems to be up now, can you try re-triggering the sanity by pushing to your fork. We can merge this then.

mapuri:
@unclejack,  looks like this needs a rebase with the recent changes.

PR #40, opened 2016-01-19
use control_interface as variable for interface carrying control traffic

mapuri:
monitor_interface variable is ceph specific. This change introduces a new host variable called control_interface to depict the interface used as control interface on hosts deploying contiv services and it's dependencies like etcd, swarm, ucarp etc. control_interface replaces monitor_interface in contiv specific plays.

/cc @DivyaVavili as this may affect your netplugin installer work.

### DIFF
--- a/group_vars/all
+++ b/group_vars/all
@@ -1,17 +1,19 @@
 ---
 # Variables here are applicable to all host groups
 
+# ceph specific variables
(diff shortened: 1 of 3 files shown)


erikh:
LGTM once tests pass

PR #41, opened 2016-01-20
Tasks to bring up netplugin in ACI mode

dvavili:
The current set of changes bring up netplugin-node in ACI mode when aci_mode is set to true. As part of this, the following changes are made:
- Launch aci-gw container with respective environment variables
- Replace docker-compose with contiv version of docker-compose

### DIFF
--- a/roles/contiv_network/tasks/aci_tasks.yml
+++ b/roles/contiv_network/tasks/aci_tasks.yml
@@ -1,0 +1,28 @@
+# These tasks are run when contiv_network_mode is set to aci
+
+- name: check aci-gw container image
+  shell: docker inspect contiv/aci-gw
+  register: docker_aci_inspect_result
+  ignore_errors: yes
+  run_once: true
+
+- name: pull aci-gw container
+  shell: docker pull contiv/aci-gw
+  when: "'No such image' in docker_aci_inspect_result.stderr"
+  run_once: true
+
+- name: copy shell script for starting aci-gw
+  template: src=aci_gw.j2 dest=/usr/bin/aci_gw.sh mode=u=rwx,g=rx,o=rx
+  run_once: true
+
+- name: copy systemd units for aci-gw
+  copy: src=aci-gw.service dest=/etc/systemd/system/aci-gw.service
+  run_once: true
+
+- name: start aci-gw container
+  service: name=aci-gw state=started
+  run_once: true
+
+- name: set aci mode
+  shell: contivctl global set -nwinfra aci
+  run_once: true
--- a/roles/contiv_network/tasks/main.yml
+++ b/roles/contiv_network/tasks/main.yml
@@ -68,3 +68,6 @@
   args:
     chdir: /usr/bin/
     creates: contivctl
+
+- include: aci_tasks.yml
+  when: (run_as == "master") and (contiv_network_mode == "aci")
--- a/roles/contiv_network/templates/aci_gw.j2
+++ b/roles/contiv_network/templates/aci_gw.j2
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+usage="$0 start"
+if [ $# -ne 1 ]; then
+    echo USAGE: $usage
+    exit 1
+fi
+
+case $1 in
+start)
+    set -e
+
+    /usr/bin/docker run --net=host \
+    -e "APIC_URL: {{ apic_url }}" \
+    -e "APIC_USERNAME: {{ apic_username }}" \
+    -e "APIC_PASSWORD: {{ apic_password }}" \
+    -e "APIC_LEAF_NODE: {{ apic_leaf_nodes }}" \
+    --name=contiv-aci-gw \
+    contiv/aci-gw
+    ;;
+
+stop)
+    # don't stop on error
+    /usr/bin/docker stop contiv-aci-gw
+    /usr/bin/docker rm contiv-aci-gw
+    ;;
+
+*)
+    echo USAGE: $usage
+    exit 1
+    ;;
+esac
--- a/roles/contiv_network/templates/netplugin.j2
+++ b/roles/contiv_network/templates/netplugin.j2
@@ -1,1 +1,1 @@
-NETPLUGIN_ARGS=""
+NETPLUGIN_ARGS='-docker-plugin -vlan-if {{netplugin_if}} -vtep-ip {{node_addr}}'
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -61,3 +61,28 @@
   when: docker_compose_version.stdout != "docker-compose version 1.5.2, build 7240ff3"
   tags:
     - prebake-for-dev
+
+- name: check contiv-compose version
+  shell: contiv-compose --version
+  register: contiv_compose_version
+  ignore_errors: yes
+  tags:
+    - prebake-for-dev
+
+- name: download contiv-compose
+  get_url:
+    validate_certs: "{{ validate_certs }}"
+    url: https://cisco.box.com/shared/static/eagplmveetdmwn7u2uwadna07l1m8t3s.bz2
+    dest: /tmp/contiv-compose.bz2
+  when: docker_compose_version.stdout != "libcompose-cli version 0.0.0-dev (4c0c2dd)"
+  tags:
+    - prebake-for-dev
+
+- name: install contiv-compose
+  shell: tar vxjf /tmp/contiv-compose.bz2
+  args:
+    chdir: /usr/bin/
+    creates: contiv-compose
+  when: docker_compose_version.stdout != "libcompose-cli version 0.0.0-dev (4c0c2dd)"
+  tags:
+    - prebake-for-dev
--- a/roles/contiv_network/defaults/main.yml
+++ b/roles/contiv_network/defaults/main.yml
@@ -1,0 +1,7 @@
+---
+
+# Role defaults for contiv_network
+# NOTE: Role defaults have a lower priority than inventory vars.
+# Include variables which need to be overridden by inventory vars here.
+
+contiv_network_mode: "standalone" # Accepted values: standalone, aci
--- a/roles/contiv_network/files/aci-gw.service
+++ b/roles/contiv_network/files/aci-gw.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Contiv ACI gw
+After=auditd.service systemd-user-sessions.service time-sync.target docker.service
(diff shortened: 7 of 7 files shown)


jainvipin:
@DivyaVavili 
Minor: May be by `Launch` you mean `Replace`.
Not super critical: I am thinking that we call it `contiv-compose` to avoid confusion. So instead of replacing it, this might be just a copy.

dvavili:
@jainvipin: Having contiv-compose makes sense. I can make the change to not replace the official docker-compose and install it as contiv-compose. Also, do we use contiv-compose as the recommended CLI for the demo?

jainvipin:
@DivyaVavili:

> Also, do we use contiv-compose as the recommended CLI for the demo?

Yes.

erikh:
LGTM. Nice work!

roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -0,0 +1,49 @@
+# These tasks are run when aci_mode is set
+
+- name: check aci-gw container image
+  shell: docker inspect contiv/aci-gw
+  register: docker_aci_inspect_result
+  ignore_errors: yes
```
mapuri, line 6: I think this shall have `run_once` as well.

Also I am not sure if you already explored this but having `run_once` at the include level might be cleaner Also I think this is coming mostly from the limitation of number of supported netmaster instances (And accordingly number of aci-gw instances). I am hoping we can keep this work-around limited in scope as much possible to allow easier cleanup later.

roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -0,0 +1,49 @@
...
+- name: set aci mode
+  shell: contivctl global set -nwinfra aci
+  run_once: true
+
+#NOTE: These tasks need to be modified once there is release-version for contiv/libcompose
+- name: download contiv version of docker-compose
+  get_url:
+    validate_certs: "{{ validate_certs }}"
+    url: https://cisco.box.com/shared/static/eagplmveetdmwn7u2uwadna07l1m8t3s.bz2
+    dest: /tmp/contiv-compose.bz2
```
mapuri: We should check for version before running this task.

roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -0,0 +1,49 @@
...
+
+- name: install contiv version of docker-compose
```
mapuri: would it make sense to do this part of `docker` role? I mean is this really aci specific? I think we have used contiv-compose in non aci demos as well.

roles/contiv_network/tasks/main.yml
```diff
@@ -68,3 +68,6 @@
   args:
     chdir: /usr/bin/
     creates: contivctl
+
+- include: aci_tasks.yml
+  when: (run_as == "master") and (aci_mode is defined) and (aci_mode|bool)
```
mapuri: Assuming `false` is a reasonable default for `aci_mode`, we shall define it in `group-vars/all` and default it to `false` there.

Minor but since `contiv_network` is going to have multiple deployment modes viz aci, standalone-with-vlan, standalone-with-bgp and so on. It might be better to have a string variable like `contiv-network-mode` and set it to a resonable default in `group-vars/all` along with all possible values in a comment there.

site.yml
```diff
@@ -70,6 +70,7 @@
   - { role: contiv_storage, run_as: worker }
 
 # netplugin-node hosts set up netmast/netplugin in a cluster
+# Set aci_mode to true to run node in ACI mode
```
mapuri: related to previous comment, this documentation shall be part of the networking mode variable in `group-vats/all` file

mapuri:
thanks @DivyaVavili, overall change looks in a good direction.

roles/contiv_network/tasks/aci_tasks.yml
dvavili, line 6: Yes, this task needs to have run_once. Will make the change.

But the reason I didn't include it in the include level was that we might need contiv-compose on all nodes. @jojimt @jainvipin: Let me know if this is not a requirement. I can then move the run_once variable as necessary.

roles/contiv_network/tasks/aci_tasks.yml
dvavili: As I understand, there is no release to check version against. Once it comes in, we can include the version check as noted.

roles/contiv_network/tasks/main.yml
dvavili: I did think of the default mode for aci_mode, but was not sure if it should be in group-vars/all or under contiv_network role. Was waiting for your input. :) And I'll make it contiv-network-mode to take in "standalone/aci" mode for now.

roles/contiv_network/tasks/aci_tasks.yml
mapuri: Is this built from `contiv/libcompose`?  I think it's same as `docker-compose` so it should have a version flag already.

roles/contiv_network/tasks/aci_tasks.yml
mapuri, line 6: sure, I think `contiv-compose` installation can move to `docker` role as it is just a modified version of `docker-compose`. That will make this entire file aci specific and then we can move `run_once` out.

roles/contiv_network/tasks/main.yml
mapuri: @DivyaVavili good idea, this can be a contiv_network level role variable. I like that actually as it is specific to the network role and not used otherwise.

roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -0,0 +1,34 @@
...
+  shell: docker inspect contiv/aci-gw
+  register: docker_aci_inspect_result
+  ignore_errors: yes
+  run_once: true
+
+- name: pull aci-gw container
+  shell: docker pull contiv/aci-gw
+  when: "'No such image' in docker_aci_inspect_result.stderr"
+  run_once: true
+  tags:
```
mapuri: I don't think we should prebake aci-gw container because this image will need to be released along side with netmaster/netplugin. So we need to pull the corresponding version every time contiv_network is deployed.

On related note, we should be pulling a specific version of aci-gw container here.

roles/contiv_network/templates/aci_gw.j2
```diff
@@ -0,0 +1,33 @@
...
+case $1 in
+start)
+    set -e
+
+    /usr/bin/docker run --net=host \
+    -e "APIC_URL: {{ apic_url }}" \
+    -e "APIC_USERNAME: {{ apic_username }}" \
+    -e "APIC_PASSWORD: {{ apic_password }}" \
+    -e "APIC_LEAF_NODE: {{ apic_leaf_nodes }}" \
+    --restart=unless-stopped \
```
mapuri: We need to use the `Restart` policy of systemd instead otherwise systemd will consider the service to be dead as soon as container dies and restarts as a different process

You can remove the `--restart` flag and add something on lines of the following to the unit file

```
Restart=on-failure
RestartSec=10
```

roles/contiv_network/templates/netplugin.j2
```diff
@@ -1 +1,6 @@
-NETPLUGIN_ARGS=""
+{# If an interface is specified for netplugin, use it #}
+{% if netplugin_if is defined %}
+NETPLUGIN_ARGS='-docker-plugin -vlan-if {{netplugin_if}} -vtep-ip {{node_addr}}'
+{% else %}
+NETPLUGIN_ARGS='-docker-plugin -vtep-ip {{node_addr}}'
```
mapuri: I think as we debugged this will break non-aci setups, we should just require `netplugin_if` to be defined always

Just following shall be enough:

```
NETPLUGIN_ARGS='-docker-plugin -vlan-if {{netplugin_if}} -vtep-ip {{node_addr}}'
```

site.yml
```diff
@@ -70,6 +70,7 @@
   - { role: contiv_storage, run_as: worker }
 
 # netplugin-node hosts set up netmast/netplugin in a cluster
+# Set contiv_network_mode to aci to run node in ACI mode
```
mapuri: I think this comment is not needed as we have provided possible values for `contiv_network_mode` in the vars file

roles/contiv_network/tasks/aci_tasks.yml
dvavili: Will remove prebake... As for the version, will open an issue and handle it later.

roles/contiv_network/templates/netplugin.j2
dvavili: Already opened an issue for this. https://github.com/contiv/netplugin/issues/244

mapuri: Yes I saw the issue filed in netplugin. What I meant here was that ansible needs to fail early too if `netplugin_if` is not defined as there is no good default for `netplugin_if`

The current logic will hide the problem and let's ansible move on which will beat the purpose of catching error conditions early on.

The fix for netplugin repo will help folks who are trying to start netplugin themselves, which will be rare I think once we have cluster/installer working.

dvavili: As I understand, the fix in netplugin repo will handle it in a way that it will error out gracefully. Will let @shaleman comment on this. Again, will handle it after raising an issue if required. Have other priorities to work on currently.

roles/contiv_network/tasks/aci_tasks.yml
mapuri: sgtm

roles/contiv_network/templates/netplugin.j2
mapuri: not sgtm Divya. We shouldn't put in code that we know will break and the current logic is broken.

Even after netplugin fix, as long as netplugin needs a datapath interface to be passed as argument we need to pass it correctly from ansible.

roles/docker/tasks/main.yml
```diff
@@ -61,3 +61,29 @@
   when: docker_compose_version.stdout != "docker-compose version 1.5.2, build 7240ff3"
   tags:
     - prebake-for-dev
+
+#NOTE: These tasks need to be modified once there is release-version for contiv/libcompose
```
contivbot: I think this note is not needed as once we decide to move to a new/different release of contiv/libcompose we will need to revisit this task and perhaps more likely the next tasks.

Issue #42, opened 2016-01-20
Need to bring in capabilities to reinstall/restart services


dvavili:
Encountered many issues while trying to re-run playbook on a node because of older version or after change in environment variables. Need a mechanism to identify if the services running on a node are of older version/older configuration and reinstall/restart the services as necessary.

jainvipin:
ping @erikh, @mapuri, @jojimt 

I agree with @DivyaVavili - having a way to clean-up an be able to restart will come very handy when we are in a situation that is not-repairable.

We have a choice of doing this either with ansible or a non-ansible script, as in https://github.com/contiv/netplugin/scripts/python/cleanup.py, and startPlugin.py scripts do.

Advantages of doing it in ansible:
- consistent ways of starting/restarting/stopping services
- once place for all changes
  Advantages of doing it outside:
- Assuming we'd never need this in production, this is needed primarily for stabilization cycle.

We must also answer, if this were to happen in a production environment, what is our answer? How can it be started cleanly?

erikh:
Ansible should be used to configure the system, systemd should be used to manage the services.

More direct, the python script needs to go.

mapuri:
In general, I agree that ansible can be used to configure and cleanup services with each role/specific cleanup taken care within that scope.

I am tracking the cleanup of services with #38 . I am hoping to setup/organize our roles to perform  respective cleanups with this change.

I am targeting a production use case though i.e. a node is commissioned to start up services correclty and a node is decommissioned to cleanup services. I am not sure if this will cover dev/test usecases but may be we can need to a step at a time since we will be introducing a new workflow i.e. in short-term a restart might imply run a decommission playbook followed by commission playbook.

erikh:
Sure; not suggesting it needs to happen overnight, but we should have a strategy to get rid of these ad-hoc tools that aren't resilient. The tool, which I have used before, is not very respecting of errors nor does it handle input correctly. Ansible will. That was the meat of my comment.

mapuri:
great. I think I am on same page here then. Hopefully I will have something to show and try out soon for you folks.

mapuri:
#44 addresses the service cleanup. Let's track missing items through separate issues

PR #44, opened 2016-01-21
Add cleanup.yml to perform service cleanup tasks

### LINKED ISSUE #38: need cleanup tasks when a node is decommissioned from cluster
This issue tracks identification and development of cleanup tasks that need to be executed when a node is decommissioned. A few tasks that I have in my mind are (not in any specific order of execution):
- remove the node from VIP advertisement [only master nodes]
- remove the node from etcd cluster [ only master nodes ]
- remove the node from swarm cluster [ master and worker nodes ]
- stop contiv services [ master and worker nodes ]
- reschedule application containers [ should be taken care of by swarm's rescheduling policy ?? ]

mapuri:
This one introduces cleanup tasks per service.

fixes #38 and hopefully #42 

/cc @erikh @DivyaVavili 

PS: I also added a Vagrantfile change that will let me exercise the cleanup playbook in CI.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -18,7 +18,8 @@ if ENV["http_proxy"]
 end
 
 ansible_groups = { }
-ansible_playbook = "./site.yml"
+ansible_playbook = ENV["CONTIV_ANSIBLE_PLAYBOOK"] || "./site.yml"
+ansible_tags =  ENV["CONTIV_ANSIBLE_TAGS"] || "prebake-for-dev"
 ansible_extra_vars = {
     "env" => host_env,
     "validate_certs" => "no",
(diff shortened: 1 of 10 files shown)


dvavili:
Could we check for the version of the installed services and force delete the binaries and reinstall them. #42 was filed because of issues seen because of stale binaries. Also, though I see the copying of binaries in their respective folders should overwrite them by default if the contents are different, it does not seem to happen. Would be good to identify the issue and fix that as well.

mapuri:
@DivyaVavili 

> Also, though I see the copying of binaries in their respective folders should overwrite them by default if the contents are different, it does not seem to happen. Would be good to identify the issue and fix that as well.

Yes, I think right place to take care of versions is the installation tasks. Some of our tasks still don't do good version checks, we should track with issues. Feel free to file more specific issues with each problematic binary that you hit. I will also keep an eye for missing version checks in existing tasks and fix whenever I find them.

erikh:
LGTM :tada:

PR #46, opened 2016-01-24
make python crypto workaround task more generic

mapuri:
@erikh PTAL. 

This fixes the issue that we recently addressed in volplugin (https://github.com/contiv/volplugin/pull/110) but now I am hitting in cluster-mgr while trying to pull in recent ansible there :)

Please note that I limited the fix to just making python-crypto string check more generic as the volplugin related fix has an issue with ansible installation failure under some other failure conditions as discussed here https://github.com/contiv/volplugin/pull/110#discussion_r50438273

### DIFF
--- a/roles/base/tasks/redhat_tasks.yml
+++ b/roles/base/tasks/redhat_tasks.yml
@@ -35,8 +35,8 @@
   ignore_errors: yes
 
 - name: remove python crypt egg file to work-around https://bugs.centos.org/view.php?id=9896&nbn=2
-  shell: rm -rf /usr/lib64/python2.7/site-packages/pycrypto-2.6.1-py2.7.egg-info
-  when: python_crypto_result.msg == "Error unpacking rpm package python-crypto-2.6.1-1.el7.centos.x86_64\n"
+  shell: rm -rf /usr/lib64/python2.7/site-packages/pycrypto-*.egg-info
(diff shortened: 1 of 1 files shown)


erikh:
I don't believe this is the right approach. The reason I say that is that the ansible installation is the only reason we need this library, so if it _fails_ we should make it _not fail_, not check for a specific package and remove it (Which happens to be the one ansible relies on today). These things can drift (especially when package maintainers see a conflict) and I think we'll see this again soon enough if we do it in a too specific way.

mapuri:
@erikh 

> so if it fails we should make it not fail, not check for a specific package and remove it (Which happens to be the one ansible relies on today)

The catch with this is that if ansible installation fails due to some other reason (say certificate validation failure behind proxy for lack of failure or a timeout) we would not be able to catch it and flag it. The result will be the machine will not have ansible installed and it won't be noticed until we really need ansible on that machine.

> These things can drift (especially when package maintainers see a conflict) and I think we'll see this again soon enough if we do it in a too specific way.

Yes, if I understand you correctly the drift that you mention is with changing python-crypto package versions. But I think that get's addressed by making the error check generic as done in this patch, right? Or do you feel this patch is still specific in some other way?

erikh:
Ok, I see your point; LGTM.

PR #49, opened 2016-01-28
define ceph_stable_ice_temp_path variable

mapuri:
otherwise the `ceph-install` play fails when run on hosts with ansiblev2

/cc @erikh . PTAL. I am not very sure if this is right way to address this undefined variable. For more context on the failure have a look at this failed run `http://contiv.ngrok.io/job/Ansible_CI/41/console`


erikh:
LGTM. I don't think we'll be tackling enterprise ceph for a while yet (which is what ICE is for, IIRC)

mapuri:
aha..cool..thanks for a quick response @erikh . I will merge this once the sanity passes

PR #50, opened 2016-01-28
fix docker-socket restart logic

mapuri:
We need include the condition when service was not running.


mapuri:
/cc @erikh @DivyaVavili 

@DivyaVavili : I think you were hitting this a few days back with you installer (when docker was started but docker-tcp service was not getting started). I have tested it in cluster-mgr and seems to be working fine.

PR #52, opened 2016-01-29
ansible task for bash autocompletion support for netctl cli

vvb:
- install bash-completion package via yum/apt. This is added to the existing
  base role.
- updated the netplugin tar link to the latest release. The tar now has a
  contrib/completion/bash/netctl packaged in it.
- copy the packaged netctl file to /etc/bash_completion.d This needs sudo
  access.

### DIFF
--- a/roles/contiv_network/tasks/main.yml
+++ b/roles/contiv_network/tasks/main.yml
@@ -23,11 +23,22 @@
     url: https://github.com/contiv/netplugin/releases/download/v0.1-01-28-2016.03-55-05.UTC/netplugin-v0.1-01-28-2016.03-55-05.UTC.tar.bz2
     dest: /tmp/contivnet.tar.bz2
 
+- name: ensure netplugin directory exists
+  file: path=/usr/bin/contiv/netplugin state=directory
+
 - name: install netmaster and netplugin
   shell: tar vxjf /tmp/contivnet.tar.bz2
   args:
-    chdir: /usr/bin/
+    chdir: /usr/bin/contiv/netplugin
     creates: netmaster
+
+- name: create links for netplugin binaries
+  file: src=/usr/bin/contiv/netplugin/{{ item }} dest=/usr/bin/{{ item }} state=link
+  with_items:
+    - netctl
+    - netmaster
+    - netplugin
+    - contivk8s
 
 - name: copy environment file for netplugin
   template: src=netplugin.j2 dest=/etc/default/netplugin
@@ -36,7 +47,7 @@
   copy: src=netplugin.service dest=/etc/systemd/system/netplugin.service
 
 - name: copy bash auto complete file for netctl
-  shell: sudo cp /usr/bin/contrib/completion/bash/netctl /etc/bash_completion.d/ && sudo rm -rf /usr/bin/contrib
+  file: src=/usr/bin/contiv/netplugin/contrib/completion/bash/netctl dest=/etc/bash_completion.d/netctl state=link
 
 - name: start netplugin
   shell: systemctl daemon-reload && systemctl start netplugin
--- a/roles/contiv_network/tasks/main.yml
+++ b/roles/contiv_network/tasks/main.yml
@@ -20,7 +20,7 @@
 - name: download netmaster and netplugin
   get_url:
     validate_certs: "{{ validate_certs }}"
-    url: https://github.com/contiv/netplugin/releases/download/v0.0.0-12-11-2015.20-54-40.UTC/netplugin-v0.0.0-12-11-2015.20-54-40.UTC.tar.bz2
+    url: https://github.com/contiv/netplugin/releases/download/v0.1-01-28-2016.03-55-05.UTC/netplugin-v0.1-01-28-2016.03-55-05.UTC.tar.bz2
     dest: /tmp/contivnet.tar.bz2
 
 - name: install netmaster and netplugin
@@ -34,6 +34,9 @@
 
 - name: copy systemd units for netplugin
   copy: src=netplugin.service dest=/etc/systemd/system/netplugin.service
+
+- name: copy bash auto complete file for netctl
+  shell: sudo cp /usr/bin/contrib/completion/bash/netctl /etc/bash_completion.d/ && sudo rm -rf /usr/bin/contrib
 
 - name: start netplugin
   shell: systemctl daemon-reload && systemctl start netplugin
--- a/roles/base/tasks/redhat_tasks.yml
+++ b/roles/base/tasks/redhat_tasks.yml
@@ -25,6 +25,7 @@
     - librbd1-devel
     - lshw
     - python-requests # XXX required by ceph repo, but it has a bad package on it
+    - bash-completion
 
 - name: install and start ntp
   shell: systemctl enable ntpd
--- a/roles/base/tasks/ubuntu_tasks.yml
+++ b/roles/base/tasks/ubuntu_tasks.yml
@@ -18,6 +18,7 @@
     - perl
     - librbd-dev
     - lshw
+    - bash-completion
 
 - name: add ansible apt repository (debian)
   apt_repository:
--- a/roles/contiv_network/defaults/main.yml
+++ b/roles/contiv_network/defaults/main.yml
@@ -5,3 +5,4 @@
 # Include variables which need to be overridden by inventory vars here.
 
 contiv_network_mode: "standalone" # Accepted values: standalone, aci
+netplugin_mode: "docker" # Accepted values: docker, kubernetes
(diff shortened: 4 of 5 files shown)


roles/contiv_network/tasks/main.yml
```diff
@@ -35,6 +35,9 @@
 - name: copy systemd units for netplugin
   copy: src=netplugin.service dest=/etc/systemd/system/netplugin.service
 
+- name: copy bash auto complete file for netctl
+  shell: sudo cp /usr/bin/contrib/completion/bash/netctl /etc/bash_completion.d/ && sudo rm -rf /usr/bin/contrib
```
mapuri: this is just minor:

consider using the `file` module and make this a symlink to the extracted file. It will keep all the netplugin related extracted files in one location (say `usr/bin/netplugin/`), than scattered at multiple locations with symlink placed in desired locations. I find this a bit cleaner  (easier to debug like it easier to find if a file really comes from the last extracted release etc) but totally up to you to incorporate.

mapuri:
@vijayvikrant 

changes LGTM, I think if you rebase your branch to latest master the sanity failure shall go away.

I also have a minor clarification that I think we should discuss, before addressing ansible just for netctl. Will netctl auto-completion work with `contiv/contivctl` project? If it does, I think in production we should install bash completion for contivctl instead as that is what our users should see.

vvb:
@mapuri it won't work for contivctl as of today. I am yet to explore the contivctl cli much, but it should be really easy to make it work for contivctl. I will look into it.

roles/contiv_network/tasks/main.yml
vvb: @mapuri the intention was not having any non-executables lying around in /usr/bin/ and so I explicitly do a `rm` as a part of the ansible task. The file contrib/completion/bash/netctl is sourced to the shell and should not be executed by the user directly. Was just being skeptic about getting into a scenario where we have another netctl (auto completion file needs to have the same name as the cli binary) lying in inside /usr/bin, even if in separate folder inside. If you think it should be ok, then I will make the change.

mapuri: can we have netplugin related files in the tar ball extracted as `netplugin/contrib/...`? I think we should namespace files per project as other project might also have similar directory names and it might not always be ok to delete those.

> Was just being skeptic about getting into a scenario where we have another netctl (auto completion file needs to have the same name as the cli binary) lying in inside /usr/bin, even if in separate folder inside.

I see (but how can there be two netctl binaries in /usr/bin?). 

The scenario I was worried about (and it might similar to yours), say user has netctlv2 but bash-completion is v1 (which can happen if user forgot to copy the file to /etc/bash_completion dir, after extracting the netplugin tar file). In that case it will be hard to debug a bash-completion anomaly.

On the other hand, say if we extract the netplugin related files to `usr/bin/netplugin/` dir and just create symlinks to these files instead. Now all the files are coming from same location, so as long as we extracted the netplugin tar correctly we don't need to worry about additional steps.

But I think both of our scenarios are pretty cooked up (and shall not happen as long as we assume ansible works correclty), so totally up to you if you want to make this change :)

mapuri:
> I am yet to explore the contivctl cli much, but it should be really easy to make it work for contivctl. I will look into it.

cool, sgtm. Do you want to track it as a separate issues? I am assuming at that point we will need to make the appropriate netctl changes to not install this file then.

jainvipin:
LGTM

erikh:
does this work on redhat at all?

erikh:
also, ping CI it looks like it broke

vvb:
@erikh yes, I tested it on both redhat and ubuntu.. manually running the tasks. or by adding prebake-for-dev tag to the tasks.

erikh:
awesome thanks

vvb:
@mapuri I have done the changes to consolidate the files in one path and then create links that are expected.

mapuri:
@vijayvikrant , thanks for taking care of the comments. 

LGTM

roles/contiv_network/tasks/main.yml
```diff
@@ -20,21 +20,35 @@
 - name: download netmaster and netplugin
   get_url:
     validate_certs: "{{ validate_certs }}"
-    url: https://github.com/contiv/netplugin/releases/download/v0.0.0-12-11-2015.20-54-40.UTC/netplugin-v0.0.0-12-11-2015.20-54-40.UTC.tar.bz2
+    url: https://github.com/contiv/netplugin/releases/download/v0.1-01-28-2016.03-55-05.UTC/netplugin-v0.1-01-28-2016.03-55-05.UTC.tar.bz2
```
dvavili, line 23: This release does not seem to have -docker-plugin argument to netplugin. So, this will break when netplugin is deployed. You might have to change the NETPLUGIN_ARGS in roles/contiv_network/templates/netplugin.j2

roles/contiv_network/tasks/main.yml
```diff
@@ -20,21 +20,35 @@
...
     dest: /tmp/contivnet.tar.bz2
 
+- name: ensure netplugin directory exists
+  file: path=/usr/bin/contiv/netplugin state=directory
+
 - name: install netmaster and netplugin
   shell: tar vxjf /tmp/contivnet.tar.bz2
   args:
-    chdir: /usr/bin/
+    chdir: /usr/bin/contiv/netplugin
```
dvavili, line 32: Just trying to understand why this indirection is required. Is it just to organize things better or is there any other requirement for this?

roles/contiv_network/tasks/main.yml
vvb, line 23: @DivyaVavili Will cover it using another PR rightaway. Trying to restrict changes here for bash auto completion only.

roles/contiv_network/tasks/main.yml
vvb, line 32: @DivyaVavili as per the discussion with @mapuri, yes it is to better organise our binaries/files. This will also eases up the cleanup.

roles/contiv_network/tasks/main.yml
dvavili, line 23: @vijayvikrant: The current installer script will pull in the latest changes from ansible and deploy services. If it breaks, there will be extra effort to check the error, if someone hits it in the meantime. The change is pretty simple. Change this line in "roles/contiv_network/templates/netplugin.j2" from NETPLUGIN_ARGS='-docker-plugin -vlan-if {{netplugin_if}} -vtep-ip {{node_addr}}' to NETPLUGIN_ARGS='-vlan-if {{netplugin_if}} -vtep-ip {{node_addr}}' . Will be comfortable if it is made in this PR.

roles/contiv_network/tasks/main.yml
dvavili, line 32: Sure... Thanks for the explanation...

roles/contiv_network/tasks/main.yml
jainvipin, line 23: @DivyaVavili latest netplugin has `-dcoker-plugin` as the default....

dvavili, line 23: @jainvipin: Trying to understand it better here. If the default is -docker-plugin, it should not error out even if the argument was provided right? I see that netplugin errors out when provided with this option.

jainvipin, line 23: @DivyaVavili - the option is changed to 

```
  -plugin-mode string
        plugin mode docker|kubernetes (default "docker")
```

however what used to be `dock-plugin` is now `-plugin-mode docker` which is the default when `-plugin-mode` is not specified.

vvb:
@DivyaVavili I was able to test the code change for -plugin-mode. I have checked-in the change.
@mapuri @DivyaVavili @jainvipin @shaleman this requires the netplugin_mode variable to be defined to work. since I don't see us defining the other variables for netplugin here. I have not made that change. Ideally we need role vars for contiv_network role.

mapuri:
@vijayvikrant 

thanks for taking care of the comments.

> Ideally we need role vars for contiv_network role. 

You are right that role variable is a right place to add this. We recently introduced defaults in `contiv_network/defaults/main.yml`, we can perhaps add this there (or create `contiv_network/vars/main.yml`). I am fine with either. It will help document the possible values for this variable as well. WDYT?

dvavili:
@vijayvikrant: Thanks for checking this out and making changes accordingly.
Could we keep the netplugin_mode value in defaults/main.yml. That way it can be overridden by inventory vars as well. @mapuri: Let me know if this sounds good.

mapuri:
yes, sgtm.

vvb:
@mapuri @DivyaVavili I have added the default value. PTAL

mapuri:
thanks @vijayvikrant 

LGTM

PR #54, opened 2016-01-29
ansible: fix new ansible in volplugin

erikh:
cc @mapuri

### DIFF
--- a/roles/ceph-common/tasks/checks/check_system.yml
+++ b/roles/ceph-common/tasks/checks/check_system.yml
@@ -26,8 +26,3 @@
   changed_when: false
   sudo: false
   register: ansible_version
-
-- name: fail on unsupported ansible version
-  fail:
-    msg: "Ansible version must be >= 1.9, please update!"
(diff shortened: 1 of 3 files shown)


roles/ceph-common/tasks/checks/check_system.yml
```diff
@@ -26,8 +26,3 @@
   changed_when: false
   sudo: false
   register: ansible_version
-
-- name: fail on unsupported ansible version
-  fail:
-    msg: "Ansible version must be >= 1.9, please update!"
-  when: "{{ ansible_version.stdout | version_compare('1.9', '<') }}"
```
mapuri, line 33: @erikh 

just wondering, I think if we get rid of the curly braces this might work. Do you really want to remove the task itself?

mapuri:
LGTM, one minor clarification otherwise good to merge once sanity passes.

roles/ceph-common/tasks/checks/check_system.yml
erikh, line 33: yeah, let's just gut it.

PR #57, opened 2016-01-30
Fixing aci service and mode setup

dvavili:
Fixing service template for aci
Adding netctl command to setup aci mode as contivctl is not updated to take global commands

NOTE: Please merge after PR #52 as this requires the latest netplugin release

### DIFF
--- a/roles/contiv_network/tasks/aci_tasks.yml
+++ b/roles/contiv_network/tasks/aci_tasks.yml
@@ -24,5 +24,5 @@
   run_once: true
 
 - name: set aci mode
-  shell: contivctl global set -nwinfra aci
+  shell: contivctl net global set --fabric-mode aci
   run_once: true
--- a/roles/contiv_network/tasks/main.yml
+++ b/roles/contiv_network/tasks/main.yml
@@ -73,12 +73,12 @@
 - name: download contivctl
   get_url:
     validate_certs: "{{ validate_certs }}"
-    url: https://github.com/contiv/contivctl/releases/download/v0.0.0-01-14-2016.01-27-38.UTC/contivctl-v0.0.0-01-14-2016.01-27-38.UTC.tar.bz2
-    dest: /tmp/contivctl-v0.0.0-01-14-2016.01-27-38.UTC.tar.bz2
+    url: https://github.com/contiv/contivctl/releases/download/v0.0.0-01-31-2016.17-56-53.UTC/contivctl-v0.0.0-01-31-2016.17-56-53.UTC.tar.bz2
+    dest: /tmp/contivctl-v0.0.0-01-31-2016.17-56-53.UTC.tar.bz2
     force: no
 
 - name: install contivctl
-  shell: tar vxjf /tmp/contivctl-v0.0.0-01-14-2016.01-27-38.UTC.tar.bz2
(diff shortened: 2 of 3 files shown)


jainvipin:
LGTM
ping @jojimt

roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -23,6 +23,7 @@
   service: name=aci-gw state=started
   run_once: true
 
+# NOTE: Use Contivctl command when it is available
 - name: set aci mode
-  shell: contivctl global set -nwinfra aci
```
mapuri, line 27: Just wondering what does it take to make this command available in contivctl?

I don't think netctl is supposed to be our production cli, so we shall to avoid using it in ansible as much as possible.

erikh, line 27: just change it to `contivctl net` where `netctl` is used. that's how it's wrapped, or was.

jojimt, line 27: This looks correct.

dvavili, line 27: @erikh @mapuri: Latest contivctl net command seems to be broken/missing the global level options. ACI mode cannot be set as a result of that. I will raise an issue in contivctl repository for this. Until that is fixed, this seems to be the only way to proceed. I can merge in after the change goes in if required. But I'll leave it upto @jainvipin to take the call for the priority on this.

dvavili, line 27: @erikh : I was looking at the contivctl implementation and looks like it just points to netctl commands. Is that right or is it packaged with a separate netctl binary? So, this is what I see... https://gist.github.com/DivyaVavili/66e80e9e16f9330bce6b

erikh, line 27: https://github.com/contiv/contivctl/blob/master/main.go is literally the whole program. It should be in there under `contivctl net`.

mapuri, line 27: @DivyaVavili, @erikh 

I tried digging in a bit. The current contivctl release is a bit old and doesn't include the latest netctl command additions. I tried publishing a new release but it failed (http://contiv.ngrok.io/view/Misc/job/Contivctl_Release/11/console).

The new `version` related changes in netplugin seem to depend on generated code which seem to be the root-cause of this failure. I will see if there is a way to generate that code in contivctl without taking away it's simple approach.

Please share if you have some ideas on handling this.

ping @jainvipin @shaleman .

erikh, line 27: thanks @mapuri please let me know if I can help.

mapuri, line 27: ok, I have changed the contivctl release job to do an explicit build after `go get` to allow generating the version code.

@DivyaVavili , please try the recent contivctl release and see if it helps. Hopefully we can continue to use contivctl set of commands. https://github.com/contiv/contivctl/releases/download/v0.0.0-01-31-2016.17-56-53.UTC/contivctl-v0.0.0-01-31-2016.17-56-53.UTC.tar.bz2

mapuri:
@DivyaVavili thanks for taking care of comments

LGTM, will merge this once sanity passes.

PR #62, opened 2016-02-04
changes to bring in ucp based swarm stack

mapuri:
This PR brings in the mechanism to plug different scheduler stacks (on lines of discussion in #9) in our ansible and brings in support for docker UCP.

Changes:
- added a role `scheduler_stack` with alll supported scheduler as it's conditional dependencies
- `scheduler_provider` variable determines which scheduler stack to deploy. To being with we will support native-swarm (i.e. what exists today) and ucp-swam (i.e. the ucp stack)
- added tasks for installing ucp
- added tasks for cleaning up ucp

ping @jainvipin @erikh . PTAL when you get a chance

### DIFF
--- a/roles/ucp/templates/ucp.j2
+++ b/roles/ucp/templates/ucp.j2
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+usage="$0 start|stop"
+if [ $# -ne 1 ]; then
+    echo USAGE: $usage
+    exit 1
+fi
+
+case $1 in
+start)
+    set -e
+
+    echo starting ucp as {{ run_as }} on {{ node_name }}[{{ node_addr }}]
+
+    if [[ "{{ run_as }}" == "master" ]]; then
+        out=$(/usr/bin/docker run --rm -t --name ucp \
+            -v /var/run/docker.sock:/var/run/docker.sock \
+            docker/ucp install --host-address={{ node_addr }} \
+              --image-version={{ ucp_version }})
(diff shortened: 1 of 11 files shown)


roles/ucp/templates/ucp.j2
```diff
@@ -0,0 +1,74 @@
...
+            -v /var/run/docker.sock:/var/run/docker.sock \
+            docker/ucp install --host-address={{ node_addr }} \
+              --image-version={{ ucp_version }})
+        echo ${out}
+
+        # copy out the instance ID
+        instanceId=$(echo ${out} | egrep -o 'UCP instance ID: [a-zA-Z0-9:_]*' | \
+            awk --field-separator='UCP instance ID: ' '{print $2}')
+        echo instance-id: $instanceId
+        echo ${instanceId} > {{ ucp_remote_dir }}/{{ ucp_instance_id_file }}
```
erikh: you need t o double-quote uses of variables like this in your shell scripts. it's a good practice to avoid spaces breaking your code.

mapuri: ah, ok, yes makes sense...thanks for catching this, let me make the change.

mapuri:
@erikh I have taken care of the comment, PTAL when you get a chance.

mapuri:
@erikh I am merging this now to start pulling it in cluster repo but please do let me know in case you had any other comments.

PR #64, opened 2016-02-04
Fix build2

erikh:


roles/ceph-install/tasks/main.yml
```diff
@@ -1,3 +1,5 @@
 ---
 - include: ./installs/install_on_redhat.yml
   when: ansible_os_family == 'RedHat'
+  tags:
```
mapuri, line 4: @erikh 

Instead of adding the tag here, `ceph_install` role should have been added as dependency for `dev` here https://github.com/contiv/ansible/blob/master/roles/dev/meta/main.yml#L14

I remember having added it but looks like it is missing, may be I lost it during merge. So something like this will be one way to do it in `dev` role dependencies:

```
- { role: ceph-install, tags: 'prebake-for-dev' }
```

site.yml
```diff
@@ -14,6 +14,18 @@
   - { role: base }
   - { role: serf }
   - { role: dev }
+  - { role: vagrant }
+
+- hosts: build
```
mapuri, line 19: hmm, `dev` role was needed to pre-install binaries in the packer image but with the `build` host group not using `dev` role those binaries will still need to be pulled in our packer boxes.

I think if we take care of the `ceph-install` role as described above, then the issue that will remain is to avoid installing VBox, vagrant, packer etc that `dev` role brings in but breaks box images. I had added those binaries as I was hoping that I will be able to use `devtest` role for lab servers as well, but looks like that was a bad idea :) .. I am thinking may be for now we should just get rid of vagrant, vbox and packer tasks from dev role altogether then our boxes will be built with the bare minimal set of binaries that we need. WDYT? I will share a gist about what I have in mind as well in a bit.

mapuri, line 19: here is the gist that I think might fix `dev` role for us: https://github.com/mapuri/ansible/commit/4014dd94314afc87c9299544bf559657d07ba866

erikh, line 19: What is getting pulled in that we need to keep it for?

Let's separate anything out that's not necessary and ensure it doesn't 
get into build, e.g. they can't even use packer or vbox.

-Erik

mapuri, line 19: @erikh 

> What is getting pulled in that we need to keep it for?

things like etcd, ovs, docker, docker-compose, skydns and collins container images etc come from `dev` role that we need in packer images.

> Let's separate anything out that's not necessary and ensure it doesn't get into build, e.g. they can't even use packer or vbox.

yep, that is what I meant too. In the gist/commit that I shared I have commented out the tasks that will install vagrant, packer and vxox. I think we can move them out for now to get our build images sane. I can polish it a bit and submit a patch if you think this sounds ok.

erikh, line 19: sgtm

mapuri, line 19: cool, thanks @erikh . I have a patch (https://github.com/mapuri/ansible/commit/0f1d8da5e4ae922fdb4b23f1b235e347edf69867) that I am trying out with your virtualbox guest-additions fix.

I will send a PR in ansible and build once I have it tested out a bit.

erikh, line 19: I think you're still going to need the `build` host group to isolate `devtest` from vagrant deps

mapuri, line 19: I addressed it by using a different tag `prebake-for-test` for vagrant, vbox and packer tasks that way we can isolate build images (contiv/build) from lab hosts (contiv/lab) by using these tags differently.

mapuri:
closing this as it has been addressed by #68 and #69

PR #65, opened 2016-02-04
Adding fwd_mode option to start netplugin

abhi:
The change contains addition of fwd mode option to start netplugin in routing or bridge mode.

### DIFF
--- a/roles/contiv_network/templates/netplugin.j2
+++ b/roles/contiv_network/templates/netplugin.j2
@@ -1,1 +1,1 @@
-NETPLUGIN_ARGS='-plugin-mode {{netplugin_mode}} -vlan-if {{netplugin_if}} -vtep-ip {{node_addr}}'
+NETPLUGIN_ARGS='-plugin-mode {{netplugin_mode}} -vlan-if {{netplugin_if}} -vtep-ip {{node_addr}} -fwd-mode {{fwd_mode}}'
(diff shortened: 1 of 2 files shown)


roles/contiv_network/templates/netplugin.j2
```diff
@@ -1 +1 @@
-NETPLUGIN_ARGS='-plugin-mode {{netplugin_mode}} -vlan-if {{netplugin_if}} -vtep-ip {{node_addr}}'
+NETPLUGIN_ARGS='-plugin-mode {{netplugin_mode}} -vlan-if {{netplugin_if}} -vtep-ip {{node_addr}} -fwd-mode {{fwd_mode}}'
```
mapuri, line 1: @abhinandanpb is this a recently added flag? 

Just want to make sure that the netplugin release installed (here: https://github.com/contiv/ansible/blob/master/roles/contiv_network/tasks/main.yml#L23) by current ansible supports this otherwise you will need to bump up the release number as well.

abhi, line 1: @mapuri yes thats right . Though the binary pulled https://github.com/contiv/ansible/blob/master/roles/contiv_network/tasks/main.yml#L23  would support this flag its more intended for the new set of netplugin/ofnet changes made. I will bump up the release number once a binary generated with the changes in netplugin and ofnet.

mapuri, line 1: ok, shall we wait for new netplugin release to be out before we merge this?

Also is there a good default value of `fwd_mode`? If so then we shall add that and document accepted values in `https://github.com/contiv/ansible/blob/master/roles/contiv_network/defaults/main.yml`

mapuri:
thanks @abhinandanpb 

I will merge this as I see the comments incorporated :)

PR #68, opened 2016-02-05
add ceph-install as a dev dependency and move vbox, vagrant, packer to test role

mapuri:
This change introduces the `test` role to install packages like vagrant, vbox and packer on test machines i.e. our lab hosts while letting us not install these on our packer builds (by using the `prebake-for-test` on for lab hosts)

@erikh , this is the change we were discussing in https://github.com/contiv/ansible/pull/64

The `contiv/build` box seems broken due to the guest additions but I think this can be a good cleanup to have and not install stuff we will never use in out vagrant environments.


erikh:
LGTM

mapuri:
hmm, looks like docker's apt repo is down causing docker to stay at version 1.9 on ubuntu host and UCP seems to need 1.10, which is kind of weird as it was working fine until yesterday and I haven't changed ucp version :(

I will wait and relaunch sanity again a bit later hoping docker's apt repo comes back up.

mapuri:
looks like docker 1.10 is still not available for ubuntu (or atleast apt-get update doesn't yet pull 1.10). I have removed the pre-bake-tag for ucp images for now as they are currently used/tested only in centos environment in cluster. With this the sanity is passing.

PR #69, opened 2016-02-06
add vagrant user to docker group

### LINKED ISSUE #58: add vagrant user to docker group
As part of `vagrant` role we should also add the vagrant user to docker group. This shall avoid needing to specify `sudo` for docker commands on vagrant machines

mapuri:
fixes #58


mapuri:
merging this as it is a simple fix ported from #64

PR #72, opened 2016-02-08
Fix consul URL

erikh:


vvb:
@erikh looks like Serf has also moved. update the path for serf to `https://releases.hashicorp.com/serf/0.6.4/serf_0.6.4_linux_amd64.zip` and then the CI should pass.

PR #73, opened 2016-02-09
Installing specific docker version

dvavili:
Changes made:
Identify the os family and specifically install the docker-version based on OS type
- Debian
  Tasks:
  - install docker (debian)
- Redhat
  Tasks:
  - remove docker (redhat) - This task is required as yum does not have an option to force-reinstall
  - install docker (redhat)

### DIFF
--- a/roles/docker/defaults/main.yml
+++ b/roles/docker/defaults/main.yml
@@ -1,0 +1,4 @@
+---
+# Default values for docker role
+
+docker_version: 1.9.1
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -1,9 +1,21 @@
 ---
 # This role contains tasks for configuring and starting docker service
 
-# XXX: figure a way to download a specific docker version
-- name: install docker
-  shell: creates=/usr/bin/docker curl https://get.docker.com | bash
+- name: install docker (debian)
+  shell: curl https://get.docker.com | sed 's/docker-engine/--force-yes docker-engine={{ docker_version }}-0~{{ ansible_distribution_release }}/' | bash
+  when: ansible_os_family == "Debian"
+  tags:
+    - prebake-for-dev
+
+- name: remove docker (redhat)
+  yum: name=docker-engine state=absent
+  when: ansible_os_family == "RedHat"
+  tags:
+    - prebake-for-dev
+
+- name: install docker (redhat)
+  shell: curl https://get.docker.com | sed 's/docker-engine/docker-engine-{{ docker_version }}/' | bash
+  when: ansible_os_family == "RedHat"
   tags:
     - prebake-for-dev
 


erikh:
clever! LGTM.

jainvipin:
@erikh - seems like you'd have to use 1.10 in ansible to make it work for volplugin, thus override the ansible variable...  I suppose that'd work for you.
@DivyaVavili - LGTM, will try it out

dvavili:
@erikh Thanks... 

@jainvipin Let me know if you are facing issues with this approach. If it looks good and there are no issues, could you merge it. I do not have permissions to do so.

roles/docker/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+# Default values for docker role
+
+docker_version: 1.9.1
```
mapuri, line 4: @DivyaVavili 

just a clarification, is there a reason to use 1.9.1 version now that 1.10.1 is out? I can take care of bumping up this version if you don't see any issues.

roles/docker/tasks/main.yml
```diff
@@ -1,9 +1,21 @@
...
-  shell: creates=/usr/bin/docker curl https://get.docker.com | bash
+- name: install docker (debian)
+  shell: curl https://get.docker.com | sed 's/docker-engine/--force-yes docker-engine={{ docker_version }}-0~{{ ansible_distribution_release }}/' | bash
+  when: ansible_os_family == "Debian"
+  tags:
+    - prebake-for-dev
+
+- name: remove docker (redhat)
+  yum: name=docker-engine state=absent
+  when: ansible_os_family == "RedHat"
```
mapuri, line 12: I think we should check for docker version before uninstalling it otherwise for centos this can cause a unwanted docker restart. I will file an issue and fix it, if that sounds reasonable.

mapuri:
nice work @DivyaVavili . I have a few clarifications that we can take care in separate PRs.

roles/docker/defaults/main.yml
vvb, line 4: @mapuri looks like we might have to do slightly more than just bumping up the version.. issues that I have seen so far with 1.10

(1) ubuntu apt repository for 15.04 vivid does not have 1.10. So, ansible fails. Does not look like they want to support it. 15.10 has it. But, with 15.10 Vagrant is broken.
(2) centos72 - `start docker tcp service` fails.. because of `Feb 15 18:21:02 localhost docker: time="2016-02-15T18:21:02.453314126Z" level=fatal msg="Error starting daemon: error initializing graphdriver: \"/var/lib/docker\" contains other graphdrivers: overlay; Please cleanup or explicitly choose storage driver (-s <DRIVER>)"` - Not yet sure if it is because of 1.10.

mapuri, line 4: ah, ok. `(1)` pretty much means that we can stick to 1.9.1 for now. thanks for clarifying @vvb

PR #75, opened 2016-02-11
Centos specific fixes

shaleman:
Bunch of small fixes i had to make to bring up centos scale testbed
- centos specific changes to ansible
- netplugin master/worker roles
- bunch of iptable rules to enable cluster communication


roles/contiv_network/tasks/ovs.yml
```diff
@@ -52,3 +56,11 @@
   with_items:
     - "tcp:127.0.0.1:6640"
     - "ptcp:6640"
+
+- name: disable SE linux
```
unclejack, line 60: Why is this needed? It goes against what is recommended by the distribution and it seems a bad idea in the context of running things in Docker.

roles/base/tasks/redhat_tasks.yml
```diff
@@ -3,6 +3,7 @@
     update_cache: true
     name: '*'
     state: latest
+  ignore_errors: yes
```
erikh, line 6: errors here shouldn't be ignored; if you were having trouble let's discuss why that's not working.

roles/contiv_network/tasks/ovs.yml
erikh, line 60: I think for development images it's fine, but maybe we should leave it on just to ensure we support operation under it, people are going to want it.

erikh:
use this for iptables: https://docs.ansible.com/ansible/iptables_module.html. the commands in your form will not persist over a reboot.

erikh:
actually, it doesn't persist. We should be using the resource instead anyway.

erikh:
If we could use this instead, that would be even better: http://docs.ansible.com/ansible/firewalld_module.html

roles/contiv_network/tasks/ovs.yml
```diff
@@ -13,6 +13,10 @@
   tags:
     - prebake-for-dev
 
+- name: install openvswitch (Redhat)
```
mapuri, line 16: This task might have come from merge. ovs is installed at line 10 above, so this task can be removed.

roles/contiv_network/vars/main.yml
```diff
@@ -0,0 +1,10 @@
+---
```
mapuri, line 1: this is minor but let's use the `defaults` directory (instead of `vars`) to introduce these variables. 

We depend on ability to override some of the variables in the `installer` scripts and `cluster-mgr` heavily so choice of defining variables matters. `defaults` are preferred as they hold least precedence and allow ease of overriding from command line, inventory variables etc.

roles/docker/tasks/main.yml
```diff
@@ -53,6 +53,12 @@
 - name: remove the docker key file, if any. It shall be regenerated by docker on restart
   file: name=/etc/docker/key.json  state=absent
 
+- name: Open docker remote api port
+  command: /sbin/iptables -I INPUT 1 -p tcp --dport {{ item }} -j ACCEPT -m comment --comment "docker_api"
+  with_items:
+    - "{{ docker_api_port }}"
+
+
```
mapuri, line 61: there seems an extra blank-line here after the task

roles/etcd/tasks/main.yml
```diff
@@ -21,5 +21,16 @@
 - name: copy systemd units for etcd
   copy: src=etcd.service dest=/etc/systemd/system/etcd.service
 
+- name: Open etcd client port with iptables
+  command: /sbin/iptables -I INPUT 1 -p tcp --dport {{ item }} -j ACCEPT -m comment --comment "etcd"
+#  when: "'etcd' not in iptablesrules.stdout"
```
mapuri, line 26: suggest to remove the commented code

site.yml
```diff
@@ -79,3 +79,26 @@
   - { role: docker, etcd_client_port1: 2379 }
   - { role: swarm, run_as: master }
   - { role: contiv_network, run_as: master }
+
+# Netplugin master
```
mapuri, line 83: just curious, where do we need these host-groups?

I ask because we intentionally limited `netplugin-node` to be a single host-group for bringing up small demo clusters. For bigger clusters we should be using cluster-mgr + `service-master/worker` groups.

roles/contiv_network/tasks/ovs.yml
shaleman, line 60: I understand the implications of disabling SE linux globally.
But, when SE linux is enabled, OVS can not connect to netplugin(OVS fails to establish controller connection). If anybody knows what might be causing it, i'm willing to try.

roles/base/tasks/redhat_tasks.yml
shaleman, line 6: I get following error on two out of eight machines. It fails consistently on same two machine. Manually running "yum upgrade" goes thru.. Scavenging internet, it looks like there are few ansible bugs in this area like: https://github.com/ansible/ansible/issues/12248
If one of you can help me debug the issue, we can remove this line..

```
failed: [node2] => {"changed": true, "failed": true, "rc": 1, "results": ["Loaded plugins: fastestmirror\nLoading mirror speeds from cached hostfile\n * base: centos.arvixe.com\n * epel: mirrors.mit.edu\n * extras: repo.dimenoc.com\n * updates: mirror.trouble-free.net\nResolving Dependencies\n--> Running transaction check\n---> Package docker-engine.x86_64 0:1.9.1-1.el7.centos will be updated\n---> Package docker-engine.x86_64 0:1.10.1-1.el7.centos will be an update\n---> Package docker-engine-selinux.noarch 0:1.10.0-1.el7.centos will be updated\n---> Package docker-engine-selinux.noarch 0:1.10.1-1.el7.centos will be an update\n---> Package python-requests.noarch 0:2.6.0-1.el7_1 will be updated\n---> Package python-requests.noarch 0:2.7.0-1.el7 will be an update\n--> Finished Dependency Resolution\n\nDependencies Resolved\n\n================================================================================\n Package                Arch    Version                 Repository         Size\n================================================================================\nUpdating:\n docker-engine          x86_64  1.10.1-1.el7.centos     docker-main-repo  9.6 M\n docker-engine-selinux  noarch  1.10.1-1.el7.centos     docker-main-repo   28 k\n python-requests        noarch  2.7.0-1.el7             openstack-kilo     95 k\n\nTransaction Summary\n================================================================================\nUpgrade  3 Packages\n\nTotal download size: 9.7 M\nDownloading packages:\nDelta RPMs disabled because /usr/bin/applydeltarpm not installed.\n--------------------------------------------------------------------------------\nTotal                                              831 kB/s | 9.7 MB  00:11     \nRunning transaction check\nRunning transaction test\nTransaction test succeeded\nRunning transaction\n  Updating   : docker-engine-selinux-1.10.1-1.el7.centos.noarch             1/6 \n  Updating   : docker-engine-1.10.1-1.el7.centos.x86_64                     2/6 \n  Updating   : python-requests-2.7.0-1.el7.noarch                           3/6 \nerror: unpacking of archive failed on file /usr/lib/python2.7/site-packages/requests/packages/chardet: cpio: rename\nerror: python-requests-2.7.0-1.el7.noarch: install failed\nerror: python-requests-2.6.0-1.el7_1.noarch: erase skipped\n  Cleanup    : docker-engine-1.9.1-1.el7.centos.x86_64                      4/6 \n  Cleanup    : docker-engine-selinux-1.10.0-1.el7.centos.noarch             5/6 \n  Verifying  : docker-engine-1.10.1-1.el7.centos.x86_64                     1/6 \n  Verifying  : docker-engine-selinux-1.10.1-1.el7.centos.noarch             2/6 \n  Verifying  : python-requests-2.7.0-1.el7.noarch                           3/6 \n  Verifying  : python-requests-2.6.0-1.el7_1.noarch                         4/6 \n  Verifying  : docker-engine-selinux-1.10.0-1.el7.centos.noarch             5/6 \n  Verifying  : docker-engine-1.9.1-1.el7.centos.x86_64                      6/6 \n\nUpdated:\n  docker-engine.x86_64 0:1.10.1-1.el7.centos                                    \n  docker-engine-selinux.noarch 0:1.10.1-1.el7.centos                            \n\nFailed:\n  python-requests.noarch 0:2.6.0-1.el7_1  python-requests.noarch 0:2.7.0-1.el7 \n\nComplete!\n"]}
msg: Error unpacking rpm package python-requests-2.7.0-1.el7.noarch
python-requests-2.6.0-1.el7_1.noarch was supposed to be removed but is not!

```

roles/contiv_network/tasks/ovs.yml
shaleman, line 16: Yes, didn't notice that.. I'll remove it..

roles/contiv_network/vars/main.yml
shaleman, line 1: okay.. i can move it to defaults directory..

roles/etcd/tasks/main.yml
shaleman, line 26: okay

site.yml
shaleman, line 83: I'm changing demo installer in contiv/demo/net directory to use netplugin-master and netplugin-worker host-groups.
If you make all hosts as netplugin-node, it starts swarm manager on all hosts and swarm manager fails to come up.

shaleman:
@erikh good point on iptables commands not surviving reboot.
I'll let people smarter than me to figure out how to use ansible firewall module :)
Let me remove all iptables related changes from this patch for now..

roles/base/tasks/redhat_tasks.yml
mapuri, line 6: usually I have seen this with our old ansible where we used to install `pip` and manage packages using pip. pip and yum don't go well together. Our recent ansible has moved away from pip. So one thing we can may be try is removing pip from these servers and then try the new ansible again.

site.yml
mapuri, line 83: yeah, I understand the swarm issue. I think this change itself will not solve the swarm issue as it will happen again once we have more than one node designated as master. I think one of the ways to solve the swarm issue is to not depend on the `run_as` variable and instead use something else like a swarm bootstrap variable/scenario (like we do in etcd). WDYT? let me try and see if I can come up with a gist which might make it a bit more clear.

mapuri, line 83: here is the gist for swarm fix that seems to work: https://github.com/mapuri/ansible/commit/07cc7444256ae85c6cdd0f7c08b2086a10374448

same issue exists with ucp as well. I will change that as well and may be send a combined PR.

roles/contiv_network/tasks/ovs.yml
erikh, line 60: Please don't disable this for the rest of the project. I did some brief googling but I already have my own tall plate so I can't chase this one down for you. I doubt it's complicated.

shaleman:
Closing this since iptable commands don't survive reboot

jainvipin:
@shaleman - this is very important to get right i.e. isn't it needed for bringing up bare-metal centos cluster?

If we use ansible iptables modules or some persistent ways to save iptable rules, it might work - given we are this close to resolving this. Besides that, what is the other open issues to close this PR?

mapuri:
just a minor point, I think persisting of iptables rules might be optional in a clustered environment where a node is actively managed i.e. we can use the auto-provision logic of the cluster-manager to re-program the rules when a node comes back up from reload and is added to the cluster.

mapuri:
@shaleman wdyt about my comment on no need for persisting iptable rules?

We need the iptables rules for centos environments. And I don't think persisting them is mandatory as we always prescribe a cleanup and re-run ansible in demo setups (and in cluster-mgr managed setups it can be handled by cluster-mgr) which will reinstall these rules.

I am still not sure about the SE linux change though, is it a must?

erikh:
We could implement this: https://www.centos.org/docs/5/html/Deployment_Guide-en-US/rhlcommon-chapter-0001.html

shaleman:
@mapuri Provisioning the server every time it reboots might be okay in short term. But, it may not be right longer term. When SELinux is enabled, OVS can not connect to openflow controller on port 6634(i.e. to netplugin). I'm not sure why.. Probably something to do with port ownership.

@erikh I'm yet to study SE linux. Might be good thing to enable..

mapuri:
@shaleman 

> Provisioning the server every time it reboots might be okay in short term. But, it may not be right longer term

Actually, I was thinking we may be forced to re-provision sometimes. Say a node goes down a (and was running as a `master`) but now when it comes back up it needs to be added as `worker`. So we need to re-provision it correctly to add it back to cluster. Sometimes even when role is not changing this might be needed, for instance adding node back to ectd. As long as subsequent re-provisioning is quick and adding the node back to cluster is automated, do you see any other long term issues?

PR #76, opened 2016-02-16
fix swarm and ucp provisioning

### LINKED ISSUE #67: Swarm running as master on more than one node causes issues
Currently when more than one node is provisioned with swarm running as master, swarm doesn't work as expected. We need a mechanism to run it as master on only one node. 

Currently, we are seeing this issue when netplugin-node hostgroup is used.

mapuri:
ucp and swarm have a peer/replica based HA model i.e. the `--manage` flag only needs to be run once to init the cluster. The new nodes are the added by just using the `--join` flag. This fix addresses this by introducing a `ucp(swarm)_bootstrap_node_addr` variable which shall be set to the address of the node used to bootstrap the respective cluster.

/cc @erikh @DivyaVavili @shaleman . This shall fix #67 

Note that `swarm(ucp)_bootstrap_node_addr` will need to be set to the ip-address of the first node in the installer script or vagrant file for swarm/ucp cluster to bootstrap correctly.


erikh:
I presume this works without changes to volplugin, so LGTM.

mapuri:
@erikh yeah it doesn't affect volpguing per se, just that when we vendor the new ansible in volplugin we will need to set `swarm_bootstrap_node_add` variable to the `mon0` IP address in it's Vagrantfile.

erikh:
sgtm, thanks for the info

PR #77, opened 2016-02-17
Misc fixes

### LINKED ISSUE #55: New versions of binaries not downloaded and replaced
This issue is seen for tasks where we download binary releases and install them on the server nodes.

Example scenario: 
Netplugin/netmaster binaries are downloaded from released version and installed on server-nodes
If we were to re-run the ansible playbook after running cleanup.yml playbook and then site.yml with a new release version, the new version is not downloaded and installed

Workaround:
Remove /tmp/contivnet.tar.bz2
Remove associated binaries from /usr/bin/{netplugin/netmast/netctl/contivk8s}
Rerun ansible-playbook site.yml

This issue would be seen for any such binaries/released version.

### LINKED ISSUE #60: cleanup netplugin related files/binaries before installing it
In older installation logic for netplugin we used to extract files directly to `usr/bin` but we have now moved to using links instead as part of #52 

This causes a failure if ansible is run on a host configured using old ansible.

### LINKED ISSUE #61: use defaults for storing role defaults
right now we have a mix of two ways for storing roles variables viz. `vars` dir and `defaults` dir. It would be nice to make this consistent.

Moving to `defaults` seems reasonable as these variables hold least precedence wrt inventory variables and variable passed from command line allowing us to override these easily in our code that depends on these variables.

mapuri:
- use variables for versions of contiv binaries and embed version in the downloaded file name. Also extract the tar ball based on download task's result. Fixes #55 
- move definitions for role variables to `defaults` directory. Fixes #61 
- force creation of netplugin symlinks. Fixes #60 
- made docker installation task conditional based on installed docker version. As discussed here https://github.com/contiv/ansible/pull/73#discussion_r52930435

/cc @erikh @DivyaVavili @jojimt @jainvipin @vvb


jojimt:
I think this is reasonable.

erikh:
LGTM

erikh:
however, the build seems stuck..

mapuri:
yeah, test seems to be taking long again... I will reboot the slaves as that seemed to help last time. I think vbox goes for a toss after a certain number of vm spinups

Issue #79, opened 2016-02-17
ucp | create a local fetch directory if it doesn't exist - errors out.


vvb:
When scheduler_provider = "ucp-swarm", the following error is seen 

```
TASK: [ucp | create a local fetch directory if it doesn't exist] **************
failed: [cluster-node1 -> 127.0.0.1] => {"failed": true, "parsed": false}
[sudo via ansible, key=pinebvdaownqyquvfpxaengckhfixffn] password:


FATAL: all hosts have already failed -- aborting
```

vvb:
with ansible 2.0, the error msg is clearer..

```
TASK [ucp : create a local fetch directory if it doesn't exist] ****************
fatal: [cluster-node1 -> localhost]: FAILED! => {"changed": false, "failed": true,
 "module_stderr": "sudo: a password is required\n", "module_stdout": "", "msg": "MODULE FAILURE", "parsed": false}
```

vvb:
resolved by #80

PR #80, opened 2016-02-17
do not use sudo for local fetch ucp task

vvb:
- ansible has issues while mixing up local_action and root access.
- Fix for #79 
- For this fix itself to work we need ansible 2.0.

references: https://github.com/ansible/ansible/issues/10906

### DIFF
--- a/roles/ucp/tasks/main.yml
+++ b/roles/ucp/tasks/main.yml
@@ -29,6 +29,7 @@
 
 - name: create a local fetch directory if it doesn't exist
   local_action: file path={{ ucp_local_dir }} state=directory
+  sudo: false
   when: run_as == "master"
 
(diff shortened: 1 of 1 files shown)


mapuri:
LGTM

mapuri:
thinking more, may be we should add a task that checks/validates that the ansible control machine is running ansible v2.0. WDYT?

We can track that as a separate issue may be.

vvb:
@mapuri yes, I will add the check.

mapuri:
@vvb, is this ready for merge? 

Were you planning to add ansible version check as part of this? I have filed #87 to perform some checks, may be we can address it as part of that PR.

vvb:
@mapuri that sounds better. so, this is good to merge then.

PR #82, opened 2016-02-17
move serf from devtest host-group to the new cluster-node host-group

mapuri:
serf requires a interface/netdevice on the host to start it's discovery protocol. We don't have that knowledge while provisioning a devtest host.

serf is required for auto-discovery of nodes, which is core to node management
in contiv cluster-manager work-flows. We introduce a new host-group viz. cluster-node
for provisioning a node with all base packages required to prepare the node
for management by cluster-manager. This host-group can be used to onboard a stock host
into the cluster-manager purview.

/cc @erikh


erikh:
LGTM

vvb:
LGTM

PR #84, opened 2016-02-18
roles/etcd/tasks/main.yml: bump etcd to 2.2.5

unclejack:
This bumps etcd to the current stable release.

### DIFF
--- a/roles/etcd/tasks/main.yml
+++ b/roles/etcd/tasks/main.yml
@@ -1,17 +1,17 @@
 ---
 # This role contains tasks for configuring and starting etcd service
 
-- name: download etcd v2.1.1
+- name: download etcd v2.2.5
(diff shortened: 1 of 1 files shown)


mapuri:
etcd version update might need some testing and go ahead from netplugin and volplugin projects.

ping @shaleman @erikh to get their views.

erikh:
WFM, I will test when I vendor in the latest.

mapuri:
@unclejack, did you happen to verify netplugin sanities with this change?

I think we can merge this if netplugin is ok as well.

unclejack:
@mapuri This works just fine with netplugin. I've tested with the current box and etcd 2.2.5.

mapuri:
@unclejack thanks for verifying. LGTM

PR #85, opened 2016-02-18
ansible: enable profile_tasks

### LINKED ISSUE #18: it will be nice to get time taken by each task
This can help improving the playbook by identifying longer task

We can use something like this: https://github.com/jlafon/ansible-profile

unclejack:
ansible 2.x includes the profile_tasks plugin. We can enable this using ansible.cfg. We don't need to include the plugin because ansible 2.x already ships it.

I've tested this with a tiny playbook.

Fixes #18

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,0 +1,2 @@
+[defaults]
+callback_whitelist = profile_tasks


mapuri:
Awesome! thanks for taking care of this @unclejack 

LGTM

PR #91, opened 2016-02-19
Misc fixes and enhancements

### LINKED ISSUE #89: fix contivctl task to extract the binaries if a new tar ball was downloaded
right now the extraction task checks for binary to exist, but it should check for a new tarball to have been downloaded

### LINKED ISSUE #90: cleanup tasks still refer to vars directory
we have moved all the variable definitions to defaults but cleanup playbook is still picking up the etcd vars from vars directory which causes it to fail

mapuri:
- Update cleanup playbook to use etcd variables from default dir. fixes #90 
- fix/cleanup various download/install tasks to use the result of `get_url` task. fixes: #89 

/cc @erikh @vvb

### DIFF
--- a/cleanup.yml
+++ b/cleanup.yml
@@ -6,11 +6,9 @@
 - hosts: all
   sudo: true
   tasks:
-    - include_vars: roles/{{ item }}/vars/main.yml
-      with_items:
-      - "etcd"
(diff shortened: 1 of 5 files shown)


mapuri:
merging this as these fixes are needed to bring up installer CI

Issue #88, opened 2016-02-19
ucp/swarm replica needs to be set for HA


mapuri:
we need to set `--replica` and `--replication` flags respectively on master nodes for ucp and swarm respectively for a HA cluster.

vvb:
resolved for ucp. swarm HA is the pending bit

Issue #86, opened 2016-02-18
start docker tcp socket service fails on subsequent tries.


vvb:
steps:

In contiv/cluster repo,
`make demo`
`vagrant provision`

```
TASK [docker : start docker tcp socket service] ********************************
fatal: [cluster-node1]: FAILED! => {"changed": true, "cmd": "sudo systemctl daemon-reload && sudo systemctl stop docker && sudo systemctl start docker-tcp.socket && sudo systemctl start docker", "delta": "0:01:32.122085", "end": "2016-02-18 19:24:42.840007", "failed": true, "rc": 1, "start": "2016-02-18 19:23:10.717922", "stderr": "Warning: Stopping docker.service, but it can still be activated by:\n  docker-tcp.socket\nJob for docker.service failed because the control process exited with error code. See \"systemctl status docker.service\" and \"journalctl -xe\" for details.", "stdout": "", "stdout_lines": [], "warnings": ["Consider using 'become', 'become_method', and 'become_user' rather than running sudo"]}
```

logs:

```
[vagrant@cluster-node1 ~]$ sudo systemctl status docker.service -ln 1000
● docker.service - Docker Application Container Engine
   Loaded: loaded (/usr/lib/systemd/system/docker.service; disabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/docker.service.d
           └─env.conf
   Active: failed (Result: exit-code) since Thu 2016-02-18 19:24:42 UTC; 1min 55s ago
     Docs: https://docs.docker.com
  Process: 11387 ExecStart=/usr/bin/docker daemon -s overlay -H fd:// --cluster-store=etcd://localhost:2379 (code=exited, status=2)
 Main PID: 11387 (code=exited, status=2)

Feb 18 19:23:12 cluster-node1 systemd[1]: Starting Docker Application Container Engine...
Feb 18 19:23:12 cluster-node1 docker[11387]: time="2016-02-18T19:23:12.733787480Z" level=info msg="Graph migration to content-addressability took 0.00 seconds"
Feb 18 19:23:12 cluster-node1 docker[11387]: time="2016-02-18T19:23:12.747448331Z" level=info msg="Firewalld running: false"
Feb 18 19:24:25 cluster-node1 docker[11387]: time="2016-02-18T19:24:25.105628230Z" level=info msg="Default bridge (docker0) is assigned with an IP address 172.17.0.0/16. Daemon option --bip can be used to set a preferred IP address"
[...]

Issue #81, opened 2016-02-17
wait for ucp files to be created times out..


vvb:
when scheduler_provider = "ucp-swam"
This timeout is seen on many runs.

```
TASK [ucp : wait for ucp files to be created, which ensures the service has started] ***



failed: [cluster-node1] => (item=ucp-fingerprint) => {"elapsed": 300, "failed": true, "item": "ucp-fingerprint", "msg": "Timeout when waiting for file /tmp/ucp-fingerprint"}
failed: [cluster-node1] => (item=ucp-instance-id) => {"elapsed": 300, "failed": true, "item": "ucp-instance-id", "msg": "Timeout when waiting for file /tmp/ucp-instance-id"}

PLAY RECAP *********************************************************************
cluster-node1              : ok=132  changed=22   unreachable=0    failed=1
```

vvb:
resolved by #93

PR #93, opened 2016-02-21
ucp fixes; ucp ha support; misc fixes

vvb:
- added `-ctrl-ip {{node_addr}}` to `NETPLUGIN_ARGS` - without it, the netplugin registers with etcd with the localIP instead of the node_addr for non-etcdcluster nodes
- fix for "wait for ucp files to be created times out.. #81"
- fix for "start docker tcp socket service fails on subsequent tries. #86"
- add `--replica` option to the nodes which do ucp join. Fix for "ucp/swarm replica needs to be set for HA #88"
- make sure that docker ucp containers are removed on a `systemctl ucp stop`


roles/ucp/templates/ucp.j2
```diff
@@ -40,12 +40,20 @@ start)
     {% else -%}
     /usr/bin/docker run --rm -t --name ucp \
         -v /var/run/docker.sock:/var/run/docker.sock \
-        -e UCP_ADMIN_USER="admin" -e UCP_ADMIN_PASSWORD="orca" \
+        -e UCP_ADMIN_USER={{ ucp_admin_user }} -e UCP_ADMIN_PASSWORD={{ ucp_admin_password }} \
         docker/ucp join --host-address={{ node_addr }} \
           --image-version={{ ucp_version }} \
           --url="https://{{ service_vip }}:443" \
+    {% if ucp_replicator1_node is defined and node_addr == ucp_replicator1_node_addr -%}
```
mapuri: @vvb, I think we can simplify this logic and make it more generic by using the condition `{{ if run_as == master }}` to append the `--replica` flag to this command as only master nodes need to be replicas.

also, would you like to make similar change to `roles/swarm/templates/swarm.j2` as well in this PR? Or we can let #88 stay open until swarm is addressed.

roles/ucp/templates/ucp.j2
```diff
@@ -62,9 +70,9 @@ stop)
     docker ps -a | grep 'ucp-' | awk '{print $1}' | xargs docker rm -v
 
     # XXX: do we need to uninstall ucp too?
-    #/usr/bin/docker run --rm -t --name ucp \
-    #    -v /var/run/docker.sock:/var/run/docker.sock \
-    #    docker/ucp uninstall --id=`cat {{ ucp_remote_dir }}/{{ ucp_instance_id_file }}` \
+    /usr/bin/docker run --rm -t --name ucp \
```
mapuri: I don't think we should enable this yet (unless it is recommended by docker folks)...from my experiments last time this seems a bit too heavy wait for service cleanup as it goes around removing the images etc forcing a fresh download next time which we si something we should avoid to keep provisioning time low.

roles/ucp/templates/ucp.j2
vvb: @mapuri sure, `run_as == master` works as long as we are sure that we want all master to be replicas.. I will make the change

roles/ucp/templates/ucp.j2
vvb: @mapuri ok, let me comment it again..

roles/ucp/templates/ucp.j2
mapuri: yes, this is true elsewhere as well..

In current implementation of ansible, only master nodes are meant to be part of the cluster (viz. ucarp, etcd, netmaster, volmaster, ceph-mon, swarm and ucp)  and worker nodes are either proxies (like ectd, swarm, ucp) or not part of active cluster membership at all (netplugin, volplugin, ucarp, ceph)...

In almost all cases, for worker nodes to move to active cluster membership they need to be reconfigured. In cluster-mgr this can be taken care of by the cluster-mgr decommission followed by commission workflow, when we have a more dynamic role assignment there.

roles/ucp/templates/ucp.j2
```diff
@@ -40,12 +40,17 @@ start)
     {% else -%}
     /usr/bin/docker run --rm -t --name ucp \
         -v /var/run/docker.sock:/var/run/docker.sock \
-        -e UCP_ADMIN_USER="admin" -e UCP_ADMIN_PASSWORD="orca" \
+        -e UCP_ADMIN_USER={{ ucp_admin_user }} -e UCP_ADMIN_PASSWORD={{ ucp_admin_password }} \
         docker/ucp join --host-address={{ node_addr }} \
           --image-version={{ ucp_version }} \
           --url="https://{{ service_vip }}:443" \
+    {% if run_as is defined and run_as == "master" -%}
```
mapuri: @vvb sorry to nit pick this but we should avoid this check `run_as is defined`.

It is good to fail if this variable is not defined than treating it to be `worker` as we depend on this variable at a lot of places and a check like this might result in introduction of wrong behavior and harder to debug situations.

vvb: @mapuri sure, makes sense.. will make the change

mapuri:
@vvb thanks for taking care of the comments! LGTM

the ubuntu based sanity seems to be consistently failing for docker installation and unfortunately we won't get much support for 15.04. I have turned off ubuntu based sanity for now as centos7 is our primary OS. Can you please rebase and retrigger sanity? I will merge this once the sanity is green.

PR #94, opened 2016-02-22
turnoff sanity for Ubuntu 15.04

mapuri:
Ubuntu 15.04 seems to be not supported with latest Docker 1.10 and higher.

And it has been resulting in sanity failure due to Docker installation itself, so turning off ansible sanity for Ubuntu until we decide to support it again in future.

FYI @contiv/members


mapuri:
merging this to unblock other PRs

PR #97, opened 2016-02-23
Additional volplugin netplugin fixes

erikh:
This covers #95 (which I will close) plus the addition of contiv-network to the volplugin-test host group.

### DIFF
--- a/roles/swarm/defaults/main.yml
+++ b/roles/swarm/defaults/main.yml
@@ -2,4 +2,4 @@
 # role variable for the swarm service
 #
 swarm_api_port: 2375
-swarm_bootstrap_node_addr: ""
+swarm_bootstrap_node_name: ""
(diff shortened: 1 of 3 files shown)


mapuri:
LGTM

I think we need similar change for ucp as well. I will create an issue for that.

I realize that bringing back contiv_network role to volplugin took some manual test and effort. Assuming `contiv_network` shall not be a primary test focus of volplugin, I am wondering if we shall use the `make demo` target of the cluster repo for demos and presentations instead? Recently @vvb did some good work to stabilize it. I am wondering if we should keep that one target stable and add more CI for it? I can look into it if it makes sense.

erikh:
the only real problem with that is that I am not familiar with that environment and have no way to ensure its sanity  (e..g, via systemtests). I'm going to add a new swarm-capable test suite soon, perhaps that'd be a good place to run that as well.

need to revisit this patch; will push something new in a little bit.

erikh:
closing this and re-submitting because `git subtree` has no force push...

PR #100, opened 2016-02-24
avoid re-install of ansible, upgrade ucp version to 1.0.0

vvb:
we try to re-install ansbile when the task runs as a part of another host-group. This commits will prevent it.
older versions of UCP are not available anymore - hence moving to 1.0.0

### DIFF
--- a/roles/base/tasks/redhat_tasks.yml
+++ b/roles/base/tasks/redhat_tasks.yml
@@ -39,7 +39,15 @@
   shell: rm -rf /usr/lib64/python2.7/site-packages/pycrypto-*.egg-info
   when: '"Error unpacking rpm package python2-crypto-" in python_crypto_result.msg'
 
+- name: download ansible binary (redhat)
+  get_url:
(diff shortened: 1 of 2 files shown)


vvb:
ping @mapuri this is required for the UCP, because of the 1.0.0 changes. PTAL.

roles/base/tasks/redhat_tasks.yml
```diff
@@ -39,7 +39,15 @@
   shell: rm -rf /usr/lib64/python2.7/site-packages/pycrypto-*.egg-info
   when: '"Error unpacking rpm package python2-crypto-" in python_crypto_result.msg'
 
+- name: download ansible binary (redhat)
```
mapuri: just a minor nit, I think task name should be `download ansible rpm (redhat)`

mapuri:
@vvb LGTM, just one minor comment.

vvb:
@mapuri took care of the comment.. Thanks.

PR #104, opened 2016-02-28
Multiple fixes

### LINKED ISSUE #88: ucp/swarm replica needs to be set for HA

### LINKED ISSUE #98: ucp bootstrap node shall use node name instead of address
this is tracking the comment https://github.com/contiv/ansible/pull/97#issuecomment-187837359 to make ucp logic behave identical to swarm as done in #97

mapuri:
- use `--replication` flag for swarm. Also note that unlike ucp, we no longer need bootstartp node logic for swarm. Fixes #88 
- move to using swarm container image
- replace `ucp_bootstrap_node_addr` with `ucp_bootstrap_node_name`. Fixes #98 
- also bumped up netplugin version to pull a netctl fix https://github.com/contiv/netplugin/pull/294

/cc @erikh @vvb

### DIFF
--- a/roles/contiv_network/defaults/main.yml
+++ b/roles/contiv_network/defaults/main.yml
@@ -8,7 +8,7 @@ contiv_network_mode: "standalone" # Accepted values: standalone, aci
 netplugin_mode: "docker" # Accepted values: docker, kubernetes
 fwd_mode: "bridge" #Accepted values: bridge , routing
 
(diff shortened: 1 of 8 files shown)


mapuri:
merging this now, to pull these in cluster and net-demo-installer repos.

PR #105, opened 2016-02-29
ansible/roles/base: ensure the latest ansible is installed.

erikh:


unclejack:
LGTM

vvb:
LGTM

mapuri:
@erikh LGTM

IIRC one concern with using the entire epel-testing repo was that it might bring in test/dev packages in prod.

I have fixed the ansible url issue in https://github.com/contiv/ansible/pull/104 but this fix is definitely better if using epel-testing here is ok.

erikh:
Let's just go with this, it keeps things simpler. We can always move it 
back if there's a problem.

mapuri:
SGTM

unclejack:
We can install the packages which are dependencies for ansible before we install that package from testing and this would only bring in ansible from testing.

PR #108, opened 2016-03-01
mon-osd separation fixes. making netplugin optional for service_worker

vvb:
1. optionally running only mons on the service-master cluster. the osds are run only on service-worker in this mode. The ceph-playbook changes in this PR, are to deal with ansible's behaviour of how it deals with conditionals at the role level. It runs every task in the role applying the conditional to it. In this specific case the "check" tasks were doing a `stat` and registering the output to a variable `socket`. When the task is skipped `socket` is not what it needs to be. There are handlers that kick-in at the end of service-master workflow that depend on `socket` to be correct. so, the correct solution is to do the stat when the handlers are fired, so that the true state is reflected.
2. Making contiv_network optional for the service-worker hostgroup.
3. Necessary defaults which do not change the current behaviour.

### DIFF
--- a/group_vars/all
+++ b/group_vars/all
@@ -25,3 +25,10 @@ validate_certs: "yes"
 # env:
 # service_vip:
 # control_interface:
+
+# should mons also run osds
+mon_osd_colocation_allowed: "true"
+
+# OSDs could be run without netplugin on them
+# osd_contiv_network_include controls this behaviour
+osd_contiv_network_include: "true"
--- a/group_vars/all
+++ b/group_vars/all
@@ -26,4 +26,4 @@ validate_certs: "yes"
 # service_vip:
 # control_interface:
 
-host_capability: "compute, storage, network"
+host_capability: "can-run-user-containers, storage"
--- a/group_vars/all
+++ b/group_vars/all
@@ -26,9 +26,4 @@ validate_certs: "yes"
 # service_vip:
 # control_interface:
 
-# should mons also run osds
-mon_osd_colocation_allowed: "true"
-
-# OSDs could be run without netplugin on them
-# osd_contiv_network_include controls this behaviour
-osd_contiv_network_include: "true"
+host_capability: "compute, storage, network"
--- a/site.yml
+++ b/site.yml
@@ -59,11 +59,11 @@
   - { role: ucarp }
   - { role: docker }
   - { role: etcd, run_as: master }
-  - { role: ceph-mon, mon_group_name: service-master }
-  - { role: ceph-osd, mon_group_name: service-master, osd_group_name: service-master, when: host_capability|match('.*storage.*') }
-  - { role: scheduler_stack, run_as: master }
-  - { role: contiv_network, run_as: master, when: host_capability|match('.*network.*') }
-  - { role: contiv_storage, run_as: master }
+#  - { role: ceph-mon, mon_group_name: service-master, when: host_capability|match('.*can-run-user-containers.*') }
+#  - { role: ceph-osd, mon_group_name: service-master, osd_group_name: service-master, when: host_capability|match('.*storage.*') }
+  - { role: scheduler_stack, run_as: master, when: host_capability|match('.*can-run-user-containers.*') }
+  - { role: contiv_network, run_as: master, when: host_capability|match('.*can-run-user-containers.*') }
+#  - { role: contiv_storage, run_as: master }
 
 # service-worker hosts correspond to cluster machines that run the worker/driver
 # logic of the infra services.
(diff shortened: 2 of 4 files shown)


group_vars/all
```diff
@@ -25,3 +25,10 @@ validate_certs: "yes"
 # env:
 # service_vip:
 # control_interface:
+
+# should mons also run osds
+mon_osd_colocation_allowed: "true"
+
+# OSDs could be run without netplugin on them
+# osd_contiv_network_include controls this behaviour
+osd_contiv_network_include: "true"
```
mapuri: I am presuming we will need multiple such variables based on host capabilities viz. compute only; storage only; with compute and storage) so instead of defining ceph specific boolean variables I feel it is better to introduce a single variable like say `host-capability` of type enum with possible values like `compute-only`, `storage-only`, `compute-and-storage` and use that in the conditionals below to include the roles in various host-group, something like `when: host_capability | match(*storage*)`

This offers the benefit of cutting down on the number of variables we need and also keeps it independent of ceph as we may support different storage types in future and same host-cpabilities will apply. WDYT?

vvb: @mapuri.. yes, consolidating under a common var would be good. But I think we should specify exclusion. So, everything is included by default, unless an exclusion is specified. for example the below would specify not to run mon and osd on the given host. It can be one large string of comma separated values.

`host-capability => “not_mon, not_osd"`
`when: host-capability | match(*not_osd*)`

mapuri: well negation is kind of hard to read... also instead of mon and osd being capabilities I prefer more generic capabilities like compute and storage..mon and osd makes it specific to ceph and we will need something similar if we were to do NFS based storage on UCS B-series servers

mapuri: just to add if we really need negation it might be better achieved using logical operator so a node with no storage capability can matched using this condition `when: ! host-capability | match(*storage*)`

vvb: @mapuri yes, I agree that negation is hard to read, but the flip side is that the user needs to specify every capability he wants on the node. 

on the topic of keeping it to storage and not getting down to mon/osd. Is there a case, a host might want to run both ceph and nfs on the same node. wondering to what level should the capability be defined - being able to do nfs based storage is a capability. being able to do ceph storage is another capability. so, do you think specifying "Storage-node" is good enough?

mapuri: > but the flip side is that the user needs to specify every capability he wants on the node. 

hmm, yeah, it is a problem if we have too many capabilities. How about we set `host-capability` to default as `compute,storage` and then let user override it to specific values? Again this is only a problem if we have too many capabilities. Another way to deal with this (when we have a lot of properties) would be to give a short-hand/pre-defined variables for common capabilities combinations. Say, for UCS B-series we can define a `common-b-series-cap` variable and set it to `storage,cap1,cap2....` and users can instead use the short-hand strings instead?

> on the topic of keeping it to storage and not getting down to mon/osd. Is there a case, a host might want to run both ceph and nfs on the same node....
>  so, do you think specifying "Storage-node" is good enough?

Franky I don't know. But from what we have in `contiv_network`, I think in a contiv-cluster it will be either ceph or nfs but not a mix, so storage should good enough. And the differentiation of starting right set of storage daemons might be better made at the `contiv_storage` role (like we do for `contiv_network` to deploy netplugin in either aci, or standalone or bgp mode). 

But to keep it extensible, may be `storage-ceph` is a better capability as you point out than just `storage`.

vvb:
@mapuri PTAL

site.yml
```diff
@@ -59,11 +59,11 @@
   - { role: ucarp }
   - { role: docker }
   - { role: etcd, run_as: master }
-  #- { role: ceph-mon, mon_group_name: service-master }
-  #- { role: ceph-osd, mon_group_name: service-master, osd_group_name: service-master }
+  - { role: ceph-mon, mon_group_name: service-master }
+  - { role: ceph-osd, mon_group_name: service-master, osd_group_name: service-master, when: host_capability|match('.*storage.*') }
   - { role: scheduler_stack, run_as: master }
-  - { role: contiv_network, run_as: master }
-  #- { role: contiv_storage, run_as: master }
```
mapuri, line 66: let's leave `ceph*` and `contiv_storage` roles commented out in these host-groups until #19 is fixed otherwise it breaks the commission/decommission workflows in `cluster/master`.

group_vars/all
```diff
@@ -25,3 +25,5 @@ validate_certs: "yes"
 # env:
 # service_vip:
 # control_interface:
+
+host_capability: "compute, storage, network"
```
mapuri: @vvb 

I still have a minor nit. I think we don't need both `network` and `compute` capabilities. I think we just need two capabilities for now, viz. to say that host provides `disk-storage` and that it `can-run-user-containers`, right? 

I refer the latter as `compute` but I realize it is not a good name. Shall just rename `compute` to something like `can-run-user-containers` if keeps it clear? 

`ceph-mon`, `contiv_storage`, `contiv_network` and `scheduler_stack` needs to run on hosts with this capability.

`ceph-osd` need to run on hosts with `storage` capability.

mapuri:
@vvb thanks for incorporating the `host_capability` comment. This LGTM other than one minor comment on the capabilities to start with.

vvb:
@mapuri PTAL. the comments are taken care of. Is this good to merge now?

mapuri:
LGTM, thanks @vvb

PR #113, opened 2016-03-09
docker: load saved images

unclejack:


erikh:
LGTM

roles/docker/tasks/main.yml
```diff
@@ -36,6 +36,15 @@
   tags:
     - prebake-for-dev
 
+- stat: path=/var/docker_images
+  register: docker_images
```
mapuri, line 40: this task can also be tagged with `prebake-for-dev`, that way we don't need the explicit `is defined` check below. I tend to avoid `is defined` checks where possible as it help tasks to fail and be caught easily in scenarios where variable might not be defined due to reasons we may not foresee.

roles/docker/tasks/main.yml
```diff
@@ -36,6 +36,15 @@
...
+
+- name:
+  shell: cd /var/docker_images; images=$(ls); for i in $images; do docker load -i $i; rm $i; done
```
mapuri, line 43: may be we should `set -e` here so that script can fail if an image load fails?

roles/docker/tasks/main.yml
erikh, line 40: I get the feeling this tag has too much power and needs to be factored out into something cleaner. the fact we have to use it everywhere smells.

that said I will make the change, let's just look soon into a different way of doing this in the future.

roles/docker/tasks/main.yml
erikh, line 43: will do

roles/docker/tasks/main.yml
mapuri, line 40: the only function this tag offers is to pre-bake things in the packer image (to lower ansible provision times in vagrant environments) while not special casing them and using same ansible. So it's function is pretty limited. If there is a better way to accomplish this I would be happy to explore that option.

erikh, line 40: Yes but this happens at provisioning time, not prebake time. It's 
unpacking the images.

-Erik

mapuri, line 40: oh, may be I got it wrong. I thought @unclejack wanted to run this task as part of packer provisioner. If not then we don't need to tag this task at all.

erikh:
Carried by #115, closing this one

PR #115, opened 2016-03-09
Load images

erikh:
Carry of #113

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -38,9 +38,11 @@
 
 - stat: path=/var/docker_images
   register: docker_images
+  tags:
+    - prebake-for-dev
 
-- name:
-  shell: cd /var/docker_images; images=$(ls); for i in $images; do docker load -i $i; rm $i; done
+- name: Import saved docker images
+  shell: set -e; cd /var/docker_images; images=$(ls); for i in $images; do docker load -i $i; rm $i; done
   when: docker_images.stat.isdir is defined and docker_images.stat.isdir
   tags:
     - prebake-for-dev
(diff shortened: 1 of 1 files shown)


roles/docker/tasks/main.yml
```diff
@@ -36,6 +36,17 @@
...
     - prebake-for-dev
 
+- stat: path=/var/docker_images
+  register: docker_images
+  tags:
+    - prebake-for-dev
+
+- name:
+  shell: set -e; cd /var/docker_images; images=$(ls); for i in $images; do docker load -i $i; rm $i; done
+  when: docker_images.stat.isdir is defined and docker_images.stat.isdir
```
mapuri, line 46: `docker_images.stat.isdir is defined` check is not required in this change.

mapuri:
@erikh LGTM, one minor comment.

Also do you want to wait on @unclejack to confirm https://github.com/contiv/ansible/pull/113#discussion_r55555236 , as that will affect if this new task should be tagged at all.

erikh:
Unfortunately, I really should push this through so I can finish this 
task today. I think it will be fine; if we need to patch it out we can 
(later today).

mapuri:
sgtm, can you incorporate the last comment on `is define` check. I think we can then merge this once sanity passes.

erikh:
already done I think; sanity is running now.

mapuri:
thanks @erikh LGTM

erikh:
removing the is_defined check failed, trying again with it back in there.

erikh:
@mapuri I'm going to merge this, please tell me what to correct and I will in a future PR, I thikn this is good enough for now

PR #118, opened 2016-03-10
Fix import

erikh:
This the fix for the importing.


roles/docker/tasks/main.yml
```diff
@@ -80,13 +86,40 @@
...
+
+- name: ensure docker is started
+  service:
+    name: docker
+    state: started
+
+- stat: path=/var/docker_images
+  register: docker_images
+  tags:
+    - prebake-for-dev
```
mapuri, line 115: @erikh based on @unclejack response on slack it does appear that this task and next one need not run during packer build. I am wondering if you could take care of removing the `prebake-for-dev` tag from these tasks in this PR?

mapuri:
@erikh the sanity failure seems related to the fact that docker daemon didn't get started as `restart docker (first time)` failed and was was ignored and the next task `ensure docker is started` was not run as it is not tagged. 

I think tagging `ensure docker is started` task with `prebake-for-dev` will resolve the sanity failure and is needed for packer builds too.

mapuri:
@erikh,  when you get a chance can please look into this one? this has been pending for a bit.

erikh:
won't prebake-for-dev ONLY be run at build time? I keep asking this question because it doesn't seem that way. this is for runtime configuration of the host, not build-time.

mapuri:
@erikh 

> won't prebake-for-dev ONLY be run at build time?

yes, that is correct. There are two comments here which probably got things mixed up.

for this comment: https://github.com/contiv/ansible/pull/118#discussion_r55724691 removing prebake-for-dev from the tasks for restoring docker images (viz. `stat` and `Import saved docker images`) makes sense as they are runtime configuration tasks.

for this comment: https://github.com/contiv/ansible/pull/118#issuecomment-194989765 adding `prebake-for-dev` to `ensure docker is started` is needed to avoid failure during build. The ansible sanity runs the playbooks in a similar way as it would be run in packer-builds that is why the sanity had failed for this PR.

erikh:
ok. I'll look into it a bit more tonight.

erikh:
Closing this one, #131 is the new hotness.

PR #120, opened 2016-03-12
remove the blanket yum upgrade task

mapuri:
a call to yum module with package name as \* causes yum to try and upgrade all the packages. This causes an unnecessary upgrade of certain packages like `docker-engine`, for which we want to control the version ourselves. It is desirable to use the appropriate state=latest or a specific version info
which each yum module call than in a single blanket task.

/cc @erikh @vvb

### DIFF
--- a/roles/base/tasks/redhat_tasks.yml
+++ b/roles/base/tasks/redhat_tasks.yml
@@ -1,17 +1,14 @@
-- name: upgrade system (redhat)
-  yum:
-    update_cache: true
-    name: '*'
-    state: latest
-
 # install epel-release first to ensure the extra packages can be installed later
 - name: install epel release package (redhat)
   yum:
     name: epel-release
+    state: latest
 
-- name: install base packages (redhat)
+- name: install/upgrade base packages (redhat)
   yum:
     name: "{{ item }}"
+    update_cache: true
+    state: latest
   with_items:
     - ntp
     - unzip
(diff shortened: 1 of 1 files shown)


erikh:
I don't know if this is a good idea. For example, our kernel in particular needs to be kept up to date, along with all the dependencies on it.

mapuri:
@erikh can we add kernel-headers or something to the base package list? Without this I notice that we upgrade docker-engine in this task (now that 1.10.3 is released) and then later we downgrade it back to older version in `docker` role. I think controlling the packages and their version that we need explicitly is better.

mapuri:
@erikh 

also just to mention (even though this is a docker issue I think) docker doesn't seem very friendly to restarts with existing containers. Most of the time restarting docker just fails if there were any containers that exist at the time docker-engine is restarted. And without this change re-running ansible always causes a docker restart.

I have added `kernel` to the list of base packages, to ensure that we keep kernel up to date. I think this helps ensures that we explicitly control the packages that we need as part of base. PTAL if this looks better.

erikh:
ok. please verify this works with volplugin before merging it, this is a big one to revert.

mapuri:
thanks @erikh.

I have verified that build, volplugin and cluster seem fine with this change.

somehow the `TestVolmasterNoGlobalConfiguration` volplugin systest fails in my setup with or without this change, so I think it is unrelated. I will look into that separately.

erikh:
it logs a failure, isn't actually a failure though.

PR #123, opened 2016-03-14
Vagrantfile: add paravirt KVM & network tweaks

unclejack:
This PR makes the same change which has been made to other Vagrantfiles from the other repositories to use the KVM paravirtprovider. The PR also makes some small recommended tweaks to speed up the network in some cases.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -43,6 +43,9 @@ Vagrant.configure(2) do |config|
         node.vm.provider "virtualbox" do |vb|
             vb.customize ['modifyvm', :id, '--memory', "4096"]
             vb.customize ["modifyvm", :id, "--cpus", "2"]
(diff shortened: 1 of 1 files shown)


mapuri:
@unclejack 

I am not familiar with the tweaks here, assuming the sanity passed so LGTM :)

Are these needed to speedup the ansible sanity?

unclejack:
@mapuri These changes help a bit with time keeping in the VM (the paravirt one) and they apply some small tweaks to the VirtualBox network configuration. This is how I've been working on my OS X setup (outside of ansible/volplugin/netplugin) and it seems to be 100% stable so far.

Should we experience any kind of problem with this, we can always revert, but I think that's not likely.

mapuri:
SGTM, thanks for explaining @unclejack

PR #124, opened 2016-03-16
cleanse base role

### LINKED ISSUE #122: prune the `base` packages to minimally needed set and move the rest to `dev` or other roles
the base package installs multiple dev packages like git, gcc, perl etc which shouldn't be installed in production environment unless they are needed for running or installing the service itself

Following is the list in `base` role that we can most likely move to `dev` role:
    - ntp
    - vim
    - curl
    - git
    - mercurial
    - gcc
    - perl

mapuri:
fixes #122 

This has been pending for a while. Basically our `base` role installs bunch of stuff that is not needed in production environment. This PR moves the tasks around to help alleviate this.
Highlight of the changes:
- moved dev packages viz. vim, git, mercurial, perl, gcc etc to `dev` role
- moved `ansible` to it's own role it now get's installed as part of `dev` role and `clusterm` role
- moved `lshw` installtion to `serf` role
- cleaned up the python-crypto work-around task as we don't need it now as we have removed pip.
- fixed a bug in Golang installation task that would prevent upgrading Golang version on a host with golang.

I have validated it with build, volplugin and cluster and seems ok.

/cc @jainvipin @erikh


mapuri:
I am going to merge this as it is a mostly agreed upon cleanup :)

Issue #36, opened 2016-01-14
etcd traffic may get dropped due to default iptable/firewalld rules on centos


mapuri:
we need to install appropriate rules to allow etcd traffic something on lines of https://github.com/kubernetes/contrib/blob/master/ansible/roles/etcd/tasks/main.yml#L39-L43

mapuri:
fixed by #126

PR #126, opened 2016-03-17
add setup and cleanup of iptables rules to allow service traffic

### LINKED ISSUE #110: ip tables rules for serf's mdns discovery seem to be filling up INPUT chain
tracking an issue that was recently noticed in volplugin where multipl iptable entries were installed for mDNS ports. The rule is installed as part of serf's service setup here https://github.com/contiv/ansible/blob/master/roles/serf/templates/serf.j2#L17

```
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp spt:5353
ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            udp dpt:5353
...
...
```

mapuri:
this adds the logic to setup iptables rules as @shaleman had proposed in #75 needed by various services. Also the cleanup of the rules is handled as part of cleanup tasks.

I have still not added the selinux specific change as that needs more investigation and can be addressed as separate PR.

With this I was able to verify serf, etcd and swarm come up fine. 

fixes #110 and #36 

/cc @erikh @unclejack @jainvipin @shaleman @DivyaVavili


erikh:
these won't persist over a reboot. consider using firewalld instead: https://fedoraproject.org/wiki/FirewallD#Static_Firewall_.28system-config-firewall.2Flokkit.29

and

http://docs.ansible.com/ansible/firewalld_module.html

mapuri:
@erikh 

yes, we had discussed the iptable/firewall rule persistence in @shaleman 's PR in comments starting here https://github.com/contiv/ansible/pull/75#issuecomment-185325215

as of now my thought is that we don't need to persist these rules as they are very specific to the configuration at the time of node commission. When a node reboots, certain provision steps need to be redone in order to correctly add it back to the cluster. So re-provisioning shall take care of reinstalling the rules too with correct set of ports etc when service is reconfigured.

erikh:
but that's broken; for internet-facing machines that means there is a window where the firewall is unavailable and should theoretically allow all traffic through, possibly causing service exposure where it's not intended.

I really think whoever implements this, should take the short amount of time it will take to learn firewalld to do this more effectively. If you would like, I can implement this in a way where there are no issues.

erikh:
Sorry if that seems offensive! It's not intended to comment on you or anything like that.

mapuri:
@erikh 

> but that's broken; for internet-facing machines that means there is a window where the firewall is unavailable and should theoretically allow all traffic through, possibly causing service exposure where it's not intended.

May be I wasn't very clear, the rules we are inserting here are just to allow the traffic for a service. Without these a service will not be able to reach the outside world or other cluster members or be reachable from outside. Which I think addresses your concern of possible service exposure without host being added to cluster correctly. 

To make sure I address your concern and bring out my understanding. Take a hypothetical case, let's say when a node A is commissioned, etcd is run on port X and we open that port on the host for etcd to work (using iptables or firewalld). Now let's say the host is brought down. And during this time etcd is reconfigured on remaining hosts to run on port Y and another service is brought up on port X. Now when the host comes back we should only open port Y on this host (assuming etcd still needs to run on node A) and not port X. Having rules persisted across reboots would actually make host receive traffic on port X which might not be desirable.

I am assuming we can do this same thing with firewalld but there as well I would not like to persist these rules across reboot as they may need to change based on service configuration.

Let me know if I missed any other advantage/scenario that firewalld will provide vs iptables for our usecase here? If so, I can look into using firewalld instead of iptables for this PR.

erikh:
if the default firewall is to block all (with the exception of ssh) then I will drop the argument. I still think these rules should persist, but it's moot if the machine is going to be provisioned each time.

mapuri:
> if the default firewall is to block all (with the exception of ssh)

yes, that is what we see in Centos distros. And I think we may not have much control the defaults as this might be customer specific too.

> it's moot if the machine is going to be provisioned each time

yes, my point as well. Right now I can't think of a better way of bringing back the machine into cluster with a static, one time configuration especially with with the variety of clustered services (datastores, swarm, ucp, our own stuff) that we have plus a possibility of the configuration change while node is down. A basic service cleanup and re-provision seems simple and much more workable. Note that this re-provisioning will not entail the heavy work of downloading packages afresh, its mostly what needs to be done to correctly setup the service.

Issue #129, opened 2016-03-17
docker role: 'check docker service state' task incorrectly fails when docker is not started


mapuri:
`check docker service state` should set `ignore_errors` to true otherwise ansible can incorrectly fail under conditions where docker service is not running

mapuri:
fixed by #131

Issue #102, opened 2016-02-25
use devicemapper instead of overlayfs in docker role


mapuri:
there is a bug with using overlayfs in docker, so we need to revert back to using devicemapper for now until this is resolved:

https://github.com/docker/docker/issues/16902

mapuri:
fixed by #131

PR #131, opened 2016-03-18
Other volplugin fixes

erikh:


site.yml
```diff
@@ -20,6 +20,7 @@
   environment: '{{ env }}'
   roles: 
   - { role: base }
+  - { role: dev, run_as: master }
```
mapuri: hey @erikh , sorry to be a stubborn about this but I don't think we should merge this in `ansible/master`. 

Let me explain, `dev` role is meant to be run only during build due to one fundamental assumption around it. Basically it is designed to include all **other** roles as it's dependencies and then it assumes that ansible is run with `prebake-for-dev` tag at command-line which filters and runs only the the tasks marked as `prebake-for-dev`.

By bringing it in volplugin we are going to end-up pulling/running a lot of unwanted tasks, which we don't want. Also this will be maintenance nightmare as now we are overloading the `dev` role to be run in more than one situation (with and without `prebake-for-dev` tag).

A simpler and cleaner approach to address what we hit in volplugin yesterday would be to instead publish a new box based on new ansible and use that instead.

roles/dev/meta/main.yml
```diff
@@ -17,7 +17,6 @@ dependencies:
 - { role: etcd }
 - { role: docker }
 - { role: swarm }
-- { role: ucp }
```
mapuri, line 20: Just inline with my comment around the purpose of `dev` role, `ucp` can continue to stay here as when `dev` role is invoked with `prebake-for-dev` it automatically filters out all the tasks that are not needed in build environment.

roles/contiv_cluster/tasks/main.yml
```diff
@@ -11,10 +11,16 @@
   tags:
     - prebake-for-dev
 
+- name: check for collins image
```
mapuri, line 14: this task will need `prebake-for-dev` tag so that it is run in build (and asnible sanity) otherwise `collins_exists` will not be registered and we will get variable undefined error

roles/swarm/tasks/main.yml
```diff
@@ -1,10 +1,15 @@
 ---
 # This role contains tasks for configuring and starting swarm service
+- name: check for swarm image
```
mapuri, line 3: same as collins version check task, this task will need `prebake-for-dev` tag so that it is run in build (and asnible sanity) otherwise `swarm_exists` will not be registered and we will get variable undefined error. I think that is the reason for sanity failure

roles/docker/tasks/main.yml
```diff
@@ -80,13 +86,42 @@
...
+  service:
+    name: docker
+    state: started
+  tags:
+    - prebake-for-dev
+
+- stat: path=/var/docker_images
+  register: docker_images
+  tags:
+    - prebake-for-dev
```
mapuri: we can remove the `prebake-for-dev` tag from this task as it is runtime configuration

roles/docker/tasks/main.yml
```diff
@@ -80,13 +86,42 @@
...
+
+- name: Import saved docker images
+  shell: set -e; cd /var/docker_images; images=$(ls); for i in $images; do docker load -i $i; rm $i; done
+  when: docker_images.stat.isdir is defined and docker_images.stat.isdir
+  tags:
+    - prebake-for-dev
```
mapuri: we can remove the `prebake-for-dev` tag from this task as it is runtime configuration

site.yml
erikh: volplugin does not work without this change applied. I have already vendored these changes and they are already merged.

Please resolve the issue out of band, I really want to do this right but this is getting time-prohibitive for things that are arguably trivial issues.

erikh:
Please implement the `prebake-for-dev` changes beyond what I have already made (presuming CI passes)  yourself please. I cannot figure out what the hell this thing is for and when to use it, perhaps this is a good indicator that this needs a serious overhaul.

site.yml
mapuri: @erikh I have verified volplugin works with the box created with the new ansible and without these changes. 

If you would like me to test and push this entire PR in a way that works with ansible I am happy to help with that. But I would not push things in a way that we have to revert in next commit the moment next box comes out.

erikh: you're misunderstanding. A new box is not going to fix my build.

mapuri:
@erikh  as discussed on slack, let me take care of the changes in this PR. I will send a separate PR in a bit and we can close this one then if all looks good.

mapuri:
@erikh I have incorporated the changes in the branch and updated this PR. I think this shall be ready to merge once sanity passes but please take a quick look in case I did something unexpected here.

Thanks again for your understanding and patience with me :)

mapuri:
fixes #102 

fixes #129

erikh:
LGTM

PR #134, opened 2016-03-22
Add user specified Bridge domain and Contract mode

saramach:


jojimt:
LGTM.
@mapuri Please merge this. We need this change for IT.

roles/contiv_network/templates/aci_gw.j2
```diff
@@ -16,6 +16,8 @@ start)
     -e "APIC_PASSWORD={{ apic_password }}" \
     -e "APIC_LEAF_NODE={{ apic_leaf_nodes }}" \
     -e "APIC_PHYS_DOMAIN={{ apic_phys_domain }}" \
+    -e "APIC_EPG_BRIDGE_DOMAIN={{ apic_epg_bridge_domain }}" \
+    -e "APIC_CONTRACTS_UNRESTRICTED_MODE={{ apic_contracts_unrestricted_mode }}" \
```
jojimt, line 20: @saramach we should specify defaults for these two new vars in roles/contiv_network/defaults/main.yml.

mapuri, line 20: Yes, if there are good defaults we should define them.

This might be obvious but here are a few things to consider when adding defaults:
- a default value shall not cause bad things to happen in a users's setup. 
- if a variable is a must have for functionality to work and there is no good default. It is perfectly ok to let them remain undefined. This helps ansible fail early and errors are caught immediately.

mapuri:
LGTM, we can merge this once we close on default values

@saramach @jojimt 

I think it would help our users if we document some of these variables. I have started a document in cluster repo here https://github.com/contiv/cluster/blob/master/management/ansible_vars.md

Would it be possible for you to add a description for the variables needed to configure netplugin in ACI mode?

jojimt:
@mapuri These vars and the aci steps are documented in detail in the demo repo.

mapuri:
SGTM.

@saramach one more minor comment, all your commits seems to be published with Divya as the author. Not sure if you want that :) you might need to setup git config of your sever.

saramach:
Yes. Fixed that :)

PR #148, opened 2016-03-31
Point contiv-compose to the updated url

jojimt:
contiv-compose has been updated to reflect the recent changes to config model. This PR makes ansible point to the correct version.


mapuri:
@jojimt LGTM, sanity failed due to an unrelated error I think. Can you push to the same branch again to retrigger the sanity. We can merge this once sanity passes.

PR #152, opened 2016-04-01
Restart aci-gw and cleanup contiv-compose on restart

jojimt:


roles/docker/tasks/cleanup.yml
```diff
@@ -12,3 +12,9 @@
   become: true
   with_items:
     - "{{ docker_api_port }}"
+
+- name: cleanup contiv-compose
```
mapuri, line 16: @jojimt this will force a download even if contiv-compose version has not changed. If we are planning to maintain contiv-compose, we shall version it just like other binaries.

jainvipin, line 16: @mapuri - you mean put the version like 'contiv-compose version' or version built into the binary name?

mapuri, line 16: yes, we already have the version check in place just that looks like we might not be bumping it up when releasing contiv-compose

https://github.com/contiv/ansible/blob/master/roles/docker/tasks/main.yml#L147

jojimt, line 16: Yes, obviously it will force a download of contiv-compose iff the user initiates a restart with the -r option. Given the priorities, I don't have the bandwidth to add versioning for compose. Meanwhile, from a user's perspective, this is broken. I think we should fix what is broken and then work on making it perfect. Alternatively, if you can submit a fix for this, I am fine with it.

jojimt, line 16: Is this the version check you are talking about? Seems pretty broken to me...

```
- name: check contiv-compose version
  shell: contiv-compose --version
  register: contiv_compose_version
  ignore_errors: yes
  tags:
    - prebake-for-dev

- name: download contiv-compose
  get_url:
    url: https://cisco.box.com/shared/static/24q94bb94nz4wfnl20afwulffh3lcyt2.bz2
    dest: /tmp/contiv-compose.bz2
  when: docker_compose_version.stdout != "libcompose-cli version 0.0.0-dev (4c0c2dd)"
  tags:
    - prebake-for-dev

- name: install contiv-compose
  shell: tar vxjf /tmp/contiv-compose.bz2
  args:
    chdir: /usr/bin/
    creates: contiv-compose
  when: docker_compose_version.stdout != "libcompose-cli version 0.0.0-dev (4c0c2dd)"
  tags:
    - prebake-for-dev
```

mapuri, line 16: can you please explain, how is it broken?

Assuming `0.0.0-dev (4c0c2dd)` can be a version-variable string and `contiv-compose --version` outputs the right version.

jojimt, line 16: At first glance, I see two problems:
1) Obvious... contiv-compose is checked against docker-compose version instead of contiv-compose.
2) Per ansible documentation http://docs.ansible.com/ansible/shell_module.html, the creates clause will cause the task to be skipped if the file exists.

Hope this helps.

mapuri, line 16: ah ok, these are bug fixes. But is the logic broken? We will still need `contiv-compose --version` to return correct version which we don't seem to be doing today.

jojimt, line 16: Sorry, those are bugs. Not "bug fixes". That's what I mean when I say it is broken :|

jainvipin, line 16: @mapuri - if we introduce contiv-compose --version, would it still not require a download to check the version? I am thinking then that we also need a release schedule to put the versioning on the image names?

mapuri, line 16: > would it still not require a download to check the version?

yes and the version check is there. It is broken as @joji pointed out but it is there and can be fixed.

> I am thinking then that we also need a release schedule to put the versioning on the image names?

no, we don't need to. We check for output of `contiv-compose --version` and match it to desired version. It will only install if the version output doesn't match the expected output.

jojimt, line 16: Just FYI: there are still more issues with this ansible. The goal of a restart is not correctly captured. Whenever we do a restart, it tries to do an os update. Not only does this consume time, but it also fails intermittently. When this happens, the whole cluster gets messed up because we have lot of tasks with the run_once option. When one node fails, it picks the next available node to run these run_once tasks (e.g.aci-gw service). However, these services do not get cleaned up upon a restart. Now you have multiple such services running where there should be only one.

mapuri:
closing this as the fixes are addressed in #153

PR #153, opened 2016-04-05
fix contiv-network issues

jainvipin:
- update aci-gw container
- update netplugin images
- new version for contiv-compose
- minor fixes: check version variable


mapuri:
LGTM

PR #155, opened 2016-04-08
Etcd related fixes

### LINKED ISSUE #154: problem with dependence on download_result
we use `download_result | changed` in a few installation tasks to only perfom installation when a binary was downloaded. This is mostly an optimization to avoid extraction of tars if a binary was not downloaded. However, it breaks the desired behavior in case of downgrades when a previous vinary might already exist and is not downloaded again. As a result `download_result | changed` evaluates to false and prevents the downgraded binary from being reinstalled.

This bug tracks cleaning up the optimization from all places. A part of this was addressed in #153

mapuri:
- made etcd version configurable through variable. I am starting with default version as 2.3.1
- made the etcd timeout values configurable and set them to some conservative values
- also fixed a issue with use of `download_result` based installation logic. Fixes #154 

/cc @erikh @shaleman  @unclejack . I am assuming volplugin ad netplugin are going to be fine with this upgrade to newer etcd version.


erikh:
LGTM

PR #157, opened 2016-04-11
ansible/roles/etcd/templates/etcd.j2: fix etcdctl commands to use non…

erikh:
…-deprecated flags


mapuri:
LGTM. thanks @erikh

PR #158, opened 2016-04-13
Update ovs.yml file

gaurav-dalvi:
Changing current IPtables rule for VXLAN port. Changing protocol from tcp to udp. Tested on centos 7.
VXLAN uses udp protocol so need to change here .


mapuri:
thanks @gaurav-dalvi . 

LGTM, but you will also need to fix the corresponding cleanup task here: https://github.com/contiv/ansible/blob/master/roles/contiv_network/tasks/ovs_cleanup.yml#L29

gaurav-dalvi:
makes sense. I should have not forgotten this one.

PR #160, opened 2016-04-13
Change in the protocol for VXLAN port, IP table rule.

gaurav-dalvi:
Changing current IPtables rule for VXLAN port. Changing protocol from tcp to udp. Tested on centos 7.
VXLAN uses udp protocol so need to change here .


mapuri:
LGTM

PR #161, opened 2016-04-13
Adding a required package libselinux-python

vvb:

Ansible workflow should work on a stock minimal centos image and libselinux-python is a required dependency that is missing in the current workflow.


mapuri:
thanks @vvb 

LGTM, can you also add corresponding Ubuntu package (here: https://github.com/contiv/ansible/blob/master/roles/base/tasks/ubuntu_tasks.yml#L10)?

vvb:
@mapuri added the package for ubuntu also.

mapuri:
thanks @vvb . LGTM

PR #162, opened 2016-04-13
automating applying ucp license

vvb:
UCP cannot be used without a valid license. So far, this was a manual step and so it was breaking automatic formation of a cluster. This PR takes care of applying the supplied UCP license automatically. This needs to be done only for the bootstrap node. A user will now have to specify the license file path when the scheduler is ucp

```
"ucp_license_file":"/home/admin/docker_subscription.lic"
```


roles/ucp/tasks/main.yml
```diff
@@ -32,6 +32,10 @@
     - "{{ ucp_instance_id_file }}"
   when: node_name != ucp_bootstrap_node_name
 
+- name: copy the ucp license to the remote machine
+  copy: src={{ ucp_license_file }} dest={{ ucp_license_remote }}
```
mapuri, line 36: @vvb shall we set the default value for `ucp_license_file` to be empty string and then conditionally run this task only when user provides a license file?

I am thinking we need to continue to support unlicensed deploy of ucp and it won't be possible with the current change as it needs `ucp_license_file` to be defined to a valid file, right?

vvb, line 36: @mapuri UCP needs a license in all cases, a trial license at minimal and so I had made it mandatory. Do you think, making it optional would be better? If so, I will make the changes.

mapuri, line 36: > UCP needs a license in all cases, a trial license at minimal and so I had made it mandatory

ah. I was under the impression that it can work without a license. Just that user will only be able to bring up just one node with it. Isn't that the case anymore? If not, then you are right the changes you have are correct.

vvb, line 36: Yes, the first node can be logged-into without applying the license, but no nodes can join the cluster etc.

mapuri, line 36: yeah, so user can continue to use it (though I am not sure what they will do with it). 

I think we should add handling for empty `ucp_license_file` name and set it to default. That way we cover the case of a single node unlicensed cluster.

vvb, line 36: ok, will update the PR with the changes..

vvb, line 36: @mapuri updated the PR to now apply the license only when `ucp_license_file` is specified and not fail otherwise.

roles/ucp/defaults/main.yml
```diff
@@ -20,3 +20,4 @@ ucp_port5: "12382"
 ucp_swarm_port: "2376"
 ucp_controller_port: "443"
 ucp_swarm_strategy: "spread"
+ucp_license_remote: "/tmp/docker_subscription.lic"
```
mapuri, line 23: minor nit: `s/ docker_subscription / ucp_subscription`

mapuri:
thanks @vvb 

LGTM

PR #164, opened 2016-04-13
update versions of contiv services, ucp and swarm

mapuri:


vvb:
@mapuri may be we should also bump up docker_version to `1.10.3`

mapuri:
yeah, the only issue is it doesn't work with Ubuntu. But I think you are right let me do that as well. I think net demo installer already set's docker version correctly based on OS type so we shall be fine there.

erikh:
@vvb 1.11 was just released.

mapuri:
ok, let me set it to 1.11

mapuri:
@vvb brought up a good point that we may need to test 1.11 with our services before setting as a default. For now let me leave it as is. I think we are anyways setting desired version per project so this not urgent.

erikh:
Makes complete sense, sorry to distract.

PR #169, opened 2016-04-19
misc fixes

vvb:
1. position of `--replica` flag in the UCP command seems to matter. The other master nodes were not coming up as replicas earlier. This fixes it.
2. Added `serf` role to `cluster-control` host-group. This is required for clusterm to run.

### DIFF
--- a/roles/ucp/templates/ucp.j2
+++ b/roles/ucp/templates/ucp.j2
@@ -49,16 +49,15 @@ start)
     /usr/bin/docker run --rm -t --name ucp \
         -v /var/run/docker.sock:/var/run/docker.sock \
         -e UCP_ADMIN_USER={{ ucp_admin_user }} -e UCP_ADMIN_PASSWORD={{ ucp_admin_password }} \
+    {% if run_as == "master" -%}
+        docker/ucp join {{ ucp_controller_replica }} --host-address={{ node_addr }} \
+    {% else -%}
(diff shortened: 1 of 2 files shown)


roles/ucp/templates/ucp.j2
```diff
@@ -49,17 +49,16 @@ start)
     /usr/bin/docker run --rm -t --name ucp \
         -v /var/run/docker.sock:/var/run/docker.sock \
         -e UCP_ADMIN_USER={{ ucp_admin_user }} -e UCP_ADMIN_PASSWORD={{ ucp_admin_password }} \
+    {% if run_as == "master" -%}
+        docker/ucp join {{ ucp_controller_replica }} --host-address={{ node_addr }} \
```
mapuri, line 53: @vvb interesting issue. Just curious, how did you validate that this was the issue? 

The order of command flags usually shouldn't  matter. Is this something to be reported to ucp folks?

mapuri:
LGTM

roles/ucp/templates/ucp.j2
vvb, line 53: @mapuri yes, so on 3 node master cluster the number of ucp infra containers were not how many they should be. uday from docker team mentioned about the positioning aspect of the --replica flag. I tried out the changes, and it works now.

PR #172, opened 2016-04-21
roles/dev/tasks: bump Go to 1.6.2

unclejack:
This small PR bumps Go to 1.6.2. This will pull in some security updates, some needed bug fixes and some potential small performance boosting changes.

### DIFF
--- a/roles/dev/tasks/os_agnostic_tasks.yml
+++ b/roles/dev/tasks/os_agnostic_tasks.yml
@@ -1,11 +1,15 @@
+- name: Set Go version
+  set_fact:
+    go_version: "1.6.2"
+
 - name: download Golang v1.6
   get_url:
     validate_certs: "{{ validate_certs }}"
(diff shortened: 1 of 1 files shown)


roles/dev/tasks/os_agnostic_tasks.yml
```diff
@@ -1,11 +1,11 @@
 - name: download Golang v1.6
   get_url:
     validate_certs: "{{ validate_certs }}"
-    url: https://storage.googleapis.com/golang/go1.6.linux-amd64.tar.gz
+    url: https://storage.googleapis.com/golang/go1.6.2.linux-amd64.tar.gz
     dest: /tmp/go1.6.linux-amd64.tar.gz
```
mapuri: the `dest` filename needs update as well. The current sanity failure is related to this.

May be we can make the go version a variable and use that everywhere, just like we do for some other roles like contiv_network, contiv_storage etc.

unclejack:
@mapuri done, PTAL

mapuri:
LGTM

PR #174, opened 2016-04-21
adding the metadata size argument to lvcreate

vvb:

The default metadata size chosen by lvcreate is very small and not sufficient.

### DIFF
--- a/roles/docker/defaults/main.yml
+++ b/roles/docker/defaults/main.yml
@@ -6,3 +6,4 @@ docker_version: 1.9.1
 docker_rule_comment: "docker api"
 docker_device: ""
 docker_device_size: "10000MB"
+docker_device_metadata_size: "1000MB"
--- a/roles/docker/tasks/create_docker_device.yml
+++ b/roles/docker/tasks/create_docker_device.yml
@@ -24,6 +24,6 @@
   ignore_errors: true
 
 - name: lvcreate contiv-dockerthin
-  shell: lvcreate -n dockerthin -T contiv --size {{ docker_device_size }}
+  shell: lvcreate -n dockerthin -T contiv --size {{ docker_device_size }} --poolmetadatasize {{ docker_device_metadata_size  }}
   when: lvcreated|failed
   register: thin_provisioned


roles/docker/defaults/main.yml
```diff
@@ -6,3 +6,4 @@ docker_version: 1.9.1
 docker_rule_comment: "docker api"
 docker_device: ""
 docker_device_size: "10000MB"
+docker_device_size: "1000MB"
```
mapuri: @vvb did you mean to define `docker_device_metadata_size: "1000MB"` here? Otherwise looks like `docker_device_size` is getting defined twice.

vvb: @mapuri ah! yes, horrors of having separate test and dev workspaces. will modify

mapuri:
thanks @vvb, changes LGTM

just for my understanding, how do we choose these defaults? Would these values need to depend on available disk on the host? Or are these values conservative enough?

Also let's wait on @erikh in case he has any comments, I am not very familiar with the lvcreate function per-se.

vvb:
@mapuri lvcreate creates a logical drive out of the given physical volume. This drive is then used by docker. Now, docker_device_size has to be lesser than the size of the passed physical volume. This is where docker saves images, container data, etc.. 10GB seems fine for now.. metadata_size ideally should be 10-20% of the docker_drive_size. I have changed the default to 1G for now. We can fine tune them later, after running the scale tests.

mapuri:
got it, makes sense. Thanks for clarifying @vvb

roles/docker/defaults/main.yml
```diff
@@ -6,3 +6,4 @@ docker_version: 1.9.1
 docker_rule_comment: "docker api"
 docker_device: ""
 docker_device_size: "10000MB"
+docker_device_metadata_size: "1000MB"
```
erikh, line 9: so, FYI: ISTR LVM accepts percentages for size paramters, so you could try this:

``` yaml
docker_device_size: "85%"
docker_device_metadata_size: "15%"
```

Which in theory should DWIW. It might be a better way of specifying this data for multiple entities on a disk than fixed sizes.

mapuri:
@vvb are you going to try out what @erikh suggested. Or is this good to merge?

vvb:
@mapuri as per the discussions that we had earlier, we are going to make sure that this be handled by the user himself and then eventually using the kickstart that we supply. I think we can leave this as-is for now.

mapuri:
cool SGTM . I will merge this then.

PR #178, opened 2016-04-26
roles/docker: bump to Docker 1.11.0

unclejack:
This bumps Docker to 1.11.0.

### DIFF
--- a/roles/docker/defaults/main.yml
+++ b/roles/docker/defaults/main.yml
@@ -2,7 +2,7 @@
 # Default values for docker role
 
 docker_api_port: 2385
-docker_version: 1.9.1
+docker_version: 1.11.0
 docker_rule_comment: "docker api"
 docker_device: ""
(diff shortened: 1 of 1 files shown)


erikh:
This changes the default, which you can do in the Vagrantfile for netplugin specifically: https://github.com/contiv/volplugin/blob/master/Vagrantfile#L59

LGTM but you should be aware that changing this here changes it for everybody.

unclejack:
@erikh The intention was to change it for everyone and include it in new boxes.

mapuri:
@unclejack, just confirming that you are saying netplugin is fine with this docker version? If so we can merge this.

erikh:
ah

unclejack:
It seems this isn't OK for netplugin.

mapuri:
@unclejack I will close this for now, let's resubmit once we have netplugin working with docker 1.11 . Let me know if you think otherwise.

PR #179, opened 2016-04-27
Nfs support + gcc

erikh:
gcc/build-essential is required to build both projects now, so it should be installed. I don't really like the argument that compilers shouldn't be on servers, so I hope we can avoid it. :)


erikh:
going to resubmit, subtree is acting weird.

PR #180, opened 2016-04-27
Nfs

erikh:
I have no idea why subtree is doing this again.

However, this patch (despite the extra commits) merges cleanly without extra stuff.

it adds nfs service support and adds gcc/build-essential to the base role.

I'd prefer to not have an argument about compilers on hosts. :)


roles/base/tasks/redhat_tasks.yml
```diff
@@ -10,6 +10,7 @@
     update_cache: true
     state: latest
   with_items:
+    - gcc
```
mapuri, line 13: just curious, are these dependencies for nfs? I have seen @jainvipin get worried when we install things like compilers in `base` role so just confirming this is really needed in `base` :)

roles/base/tasks/redhat_tasks.yml
```diff
@@ -22,3 +23,17 @@
 
 - name: install and start ntp
   service: name=ntpd state=started enabled=yes
+
+- name: enable NFS server
```
mapuri, line 27: @erikh I know this adds more work for you but we should add these tasks in a `nfs` role and add them as dependency for volplugin role (just like we have done for `ceph-install` [role here](https://github.com/contiv/ansible/blob/master/roles/contiv_storage/meta/main.yml#L2)). That way we can keep `base` role lean and keep volplugin's dependency close to it.

roles/base/tasks/redhat_tasks.yml
```diff
@@ -22,3 +23,17 @@
...
+  service:
+    name: nfs-server
+    state: started
+    enabled: yes
+  when: use_nfs
```
mapuri, line 32: We can have a more general variable like say `storage-type` with possible values as `ceph`, `nfs` or `ceph,nfs`and so on which will let us install either nfs or ceph or both based on user's deployment.

roles/base/tasks/redhat_tasks.yml
jainvipin, line 13: @mapuri - agree; we shouldn't include compile tools into user components.

erikh, line 13: What do you guys want me to do? gcc is required to build both netplugin 
and volplugin, yet the dev role will absolutely not work with the centos 
vm.

-Erik

roles/base/tasks/redhat_tasks.yml
erikh, line 32: In the unlikely event that they want both, this is more work with a 
significant drawback.

-Erik

roles/base/tasks/redhat_tasks.yml
mapuri, line 13: @erikh is this due the issue that we saw with gcc not being packaged in packer boxes? It should be fixed in new boxes (https://github.com/contiv/build/pull/60), can you try the new box `0.3.6ba5a75` ?

erikh, line 13: Oh, ok! I will try, thanks.

-Erik

roles/base/tasks/redhat_tasks.yml
erikh, line 32: I'm going to do this, sorry for carping before.

PR #181, opened 2016-04-28
ucp: support for parallel multi-master bringup

mapuri:
ucp replica and agents can be brought up only after fingerprint file has been generated by starting a master node. This kind of inhibits parallel multi-master bringup for ucp and till now we are forced to bringup the bootstrap node before other nodes.

This change adds a work-around in the ucp's unit file to wait for fingerprint file to be placed before trying to start ucp, which allows to make UCP master/replica bringup the process parallel. 

This will be useful now that we have multi-master support in cluster-manager.

I have also bumped the default version of contiv services, to save a separate PR :)

/cc @vvb


mapuri:
will merge this, seems to be working in demo cluster

PR #182, opened 2016-04-29
Nfs

erikh:
Here we go again...

New role: nfs. Role enabled in volplugin-test host group.


mapuri:
LGTM

mapuri:
@erikh the commit history seems to be leaking in from volplugin. Is it possible to manually squash the commits in `nfs` ansible branch?

erikh:
I think that's going to break in volplugin.

-Erik

mapuri:
hmmm, this messes up ansible tree's history though.

I will merge this then but I think we should discuss (may be offline) about setting up a process for contributing changes to ansible, so that it keeps the master's history clean.

erikh:
I agree, we need to have this harder to break.

erikh:
Hey, it seems the extra commits were github, not git. Witness: https://github.com/contiv/ansible/commits/master

erikh:
ah disregard, I see them now.

PR #183, opened 2016-04-29
adding logging to aci-gw container

gaurav-dalvi:
Adding python logging to flask app. It logs everything in /tmp/aci-gw-apiagent.log file. We have to map /tmp on host to /tmp on aci-gw container then we can check logs, irrespective of container is running or not.

For this, we need to change in aci_gw.j2 template to ad -v option while running aci-gw container.

### DIFF
--- a/roles/contiv_network/templates/aci_gw.j2
+++ b/roles/contiv_network/templates/aci_gw.j2
@@ -19,6 +19,7 @@ start)
     -e "APIC_EPG_BRIDGE_DOMAIN={{ apic_epg_bridge_domain }}" \
     -e "APIC_CONTRACTS_UNRESTRICTED_MODE={{ apic_contracts_unrestricted_mode }}" \
     --name=contiv-aci-gw \
+    -v /tmp:/tmp \
     contiv/aci-gw
     ;;
 


mapuri:
just wondering, is the file logging a must? Can aci_gw service log to stdout? The latter is desirable and a must imo as that way we will be able to use the standard `journalctl -xu aci_gw` to get all service logs and other benefits that system logging offers

gaurav-dalvi:
Somehow, whatever print statement is printing in apiagent.py file , I never see in journalctl logs. Thats the main reason to add these logging capabilities in this container.

mapuri:
do you see them if you run the container directly through docker? For other services that we run inside container (like collins) I do see logs in journalctl

mapuri:
the problem with using host filesystem like `/tmp` is that we are assuming it will have space and also that it's ok to loose all the logs on node reboot or service restart.

gaurav-dalvi:
I only see these type of logs in 1: docker logs 2: journalctl.

127.0.0.1 - - [29/Apr/2016 17:51:28] "POST /createAppProf HTTP/1.1" 200 -
127.0.0.1 - - [29/Apr/2016 17:55:40] "POST /deleteAppProf HTTP/1.1" 200 -
127.0.0.1 - - [29/Apr/2016 17:55:41] "POST /createAppProf HTTP/1.1" 200 -

however in apiagent.py file you can clearly see we are printing request and bunch of other stuff. Which We are not able to see.

About log persistency : 
I am not sure. as long as cluster is up , we will get those longs in real time. If node restarts, we still have log file at /tmp/ and we can at least check what was happened before node crashed or restarted.

mapuri:
hmm, can you or someone debug why other logs are not seen on the stdout/stderr? I don't think this change gels well with the way we have rest of services setup. We want consistent logging for all services.

> If node restarts, we still have log file at /tmp/ 

I am not sure if this true, I think /tmp/ is gone on node restart

gaurav-dalvi:
No need for this. Closing this PR.

PR #184, opened 2016-04-29
re-enable multi netmaster instances

mapuri:
now that netmaster supports multiple instances (https://github.com/contiv/netplugin/issues/210) we can re-enable this functionality in ansible. I think this will help netmaster's multiple instance logic get tested more often in net-demo-installer and clusterm based environments.

/cc @shaleman , please let me know if you think this will not work or see any other issues. In a basic cluster-manager demo setup, I am able to verify that multiple netmaster instances come up fine and a network create/list operation went fine.

### DIFF
--- a/roles/contiv_network/tasks/aci_tasks.yml
+++ b/roles/contiv_network/tasks/aci_tasks.yml
@@ -1,22 +1,16 @@
 # These tasks are run when contiv_network_mode is set to aci
 
-# This task could pull a specific label, but when using latest this is not needed
 - name: pull aci-gw container
   shell: docker pull contiv/aci-gw
-  run_once: true
 
 - name: copy shell script for starting aci-gw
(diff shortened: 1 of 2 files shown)


mapuri:
ping @shaleman @jainvipin 

I think we should merge this in sooner than later. Please let me know if you see any issues with enabling multi master support for netmaster in ansible.

shaleman:
@mapuri SGTM
Assuming netmaster runs only on master host groups, this looks good.

mapuri:
thanks @shaleman 

sorry I didn't want to come out as aggressive. Just thought that as we would be releasing netmaster with multimaster support, enabling it in ansible would be a good way to get it out to folks who are already trying it with installer and clustermgr.

> Assuming netmaster runs only on master host groups, this looks good.

yep, that is correct.

I will merge this then.

PR #186, opened 2016-05-05
enabling ceph to work with iptables

vvb:
The changes are made as per http://docs.ceph.com/docs/master/rados/configuration/network-config-ref/

### DIFF
--- a/roles/ceph-common/defaults/main.yml
+++ b/roles/ceph-common/defaults/main.yml
@@ -7,6 +7,10 @@
 
 fetch_directory: fetch/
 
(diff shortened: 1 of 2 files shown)


mapuri:
@vvb, changes look good. We will need corresponding cleanup rules as well.

vvb:
@mapuri Can that be in another PR? The reason is that ceph cleanup needs to be thought through, we don't have a cleanup for ceph as of today. and because storage is involved, we need to be careful.

mapuri:
@vvb sgtm, let's track ceph cleanup as a separate issue

Issue #173, opened 2016-04-21
permit/deny openvswitch_t type in selinux should run only when selinux is enabled


vvb:
permit/deny openvswitch_t type in selinux should run only when selinux is enabled

currently I see the below failure

```
Apr 21 11:58:47 ucsb-blade1 clusterm: level=info msg="TASK [contiv_network : permit openvswitch_t type in selinux] *******************"
Apr 21 11:58:51 ucsb-blade1 clusterm: level=info msg="fatal: [ucsb-blade3-FLM19145KXV]: FAILED! => {\"changed\": true, \"cmd\": \"semanage permissive -a openvswitch_t\", \"delta\": \"0:00:04.046924\", \"end\": \"2016-04-21 11:58:51.721902\", \"failed\": true, \"rc\": 1, \"start\": \"2016-04-21 11:58:47.674978\", \"stderr\": \"libsepol.link_modules: Tried to link in a non-MLS module with an MLS base. (No such file or directory).\\nlibsemanage.semanage_link_sandbox: Link packages failed (No such file or directory).\\nValueError: Could not commit semanage transaction\", \"stdout\": \"\", \"stdout_lines\": [], \"warnings\": []}"
```

when i execute is manually on the shell,

```
[admin@ucsb-blade2 ~]$ semanage permissive -a openvswitch_t
ValueError: SELinux policy is not managed or store cannot be accessed.
[admin@ucsb-blade2 ~]$
```

mapuri:
@vvb if it's easier to verify, can you verify that above issue goes away if you enable selinux on the box? Somehow the ansible's shell error doesn't match the error that you hit with manual command run. May be there is more, like we may need to set the ansible shell as well?

vvb:
@mapuri will update on this later today. I had modified my setup and ran into some issues

gaurav-dalvi:
@mapuri and @vvb :
I don't see this problem. I have two nodes (centos 7) with swarm setup.

TASK [contiv_network : permit openvswitch_t type in selinux] *******************
changed: [node2]
changed: [node1]

This is when I have selinux is disabled on both nodes:
admin@netmaster ansible]$ sestatus
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   permissive                <------------------ Disabled
Mode from config file:          enforcing
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
[...]

PR #187, opened 2016-05-05
Selinux fix

gaurav-dalvi:
Fix for the issue mentioned here : 
https://github.com/contiv/ansible/issues/173

### DIFF
--- a/roles/contiv_network/tasks/ovs.yml
+++ b/roles/contiv_network/tasks/ovs.yml
@@ -53,9 +53,15 @@
     - "tcp:127.0.0.1:6640"
     - "ptcp:6640"
 
+- name: check selinux status
+  shell: sestatus
+  register: selinux_set
+  when: (ansible_os_family == "RedHat")
+
 - name: permit openvswitch_t type in selinux
   shell: >
       semanage permissive -a openvswitch_t
+  when: (ansible_os_family == "RedHat") and (selinux_set.stdout | search(".*SELinux status:.*enabled.*"))
   become: true
 
 - name: setup iptables for vxlan vtep port


roles/contiv_network/tasks/ovs.yml
```diff
@@ -53,9 +53,16 @@
     - "tcp:127.0.0.1:6640"
     - "ptcp:6640"
 
+- name: check selinux status
+  shell: sestatus | grep "disabled"
```
mapuri: @gaurav-dalvi 

On my machine I see enabled at two lines in the command output, I think this check needs to be more specific

Can we try something like this:

```
shell: sestatus | egrep "SELinux status:.*enabled"
```

```
[vagrant@cluster-node1 ~]$ sestatus
SELinux status:                 enabled            <<<<<
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   permissive
Mode from config file:          permissive
Policy MLS status:              enabled             <<<<<<<
Policy deny_unknown status:     allowed
Max kernel policy version:      28 
```

roles/contiv_network/tasks/ovs.yml
```diff
@@ -53,9 +53,16 @@
...
+  register: no_selinux
+  when: (ansible_os_family == "RedHat")
+  ignore_errors: yes
+
 - name: permit openvswitch_t type in selinux
   shell: >
       semanage permissive -a openvswitch_t
+  when: (ansible_os_family == "RedHat") and not (no_selinux.stdout)
```
mapuri: If we make the above change, we can also change the check as `no_selinux.stdout|match( "SELinux status:.*enabled"`. This might be a bit more readable as well.

roles/contiv_network/tasks/ovs_cleanup.yml
```diff
@@ -25,9 +25,16 @@
 
 - debug: var=ports
 
+- name: check selinux status
```
mapuri: we can skip the task and check below in cleanup as cleanup errors are ignored.

roles/contiv_network/tasks/ovs.yml
gaurav-dalvi: Good Point. I will update PR with this.

roles/contiv_network/tasks/ovs.yml
gaurav-dalvi: Sounds good !

roles/contiv_network/tasks/ovs_cleanup.yml
gaurav-dalvi: Reason is to undo whatever we do in ovs.yml file here. How come removing this task, will execute "deny openvswitch_t type in selinux" for redhat when Selinix us enabled ?

mapuri: sorry, may be I wasn't clear. I just meant that we don't need to make the selinux specific cleanup ( i.e. `deny openvswitch_t type in selinux` task) conditional. We can always run it, even if it fails the error is ignored. This reduces a task and simplifies the play.

erikh: +1

gaurav-dalvi: Got it. Thanks. Anyways we have check for redhat so cleanup task only run for centos. I will remove check selinux status task from ovs_cleanup.yml file

mapuri: I think we can get rid of centos check as well in the cleanup.

roles/contiv_network/tasks/ovs_cleanup.yml
```diff
@@ -28,6 +28,7 @@
 - name: deny openvswitch_t type in selinux
   shell: >
       semanage permissive -d openvswitch_t
+  when: (ansible_os_family == "RedHat")
```
mapuri: @gaurav-dalvi sorry to nit pick. I think we can get rid of this conditional as well.

gaurav-dalvi: @mapuri : why ? It will not run on ubuntu , resulting that task will skipped as opposed to the current logic : It will still run even if its ubuntu and fail.

mapuri: > why ?

this is to be consistent with rest of the cleanup steps. We don't need conditionals in cleanup tasks unless we absolutely need them.

gaurav-dalvi: Done !

roles/contiv_network/tasks/ovs.yml
```diff
@@ -53,9 +53,16 @@
     - "tcp:127.0.0.1:6640"
     - "ptcp:6640"
 
+- name: check selinux status
+  shell: sestatus | egrep "SELinux status:.*enabled"
+  register: selinux_set
+  when: (ansible_os_family == "RedHat")
```
mapuri, line 59: @gaurav-dalvi , please kick me if you want :)

Less code brings out one more readability enhancement that we shall do here.

I think we can simplify the task further by removing the `(ansible_os_family == "RedHat")` conditional altogether from this task and next task.

mapuri:
@gaurav-dalvi also we need to squash the commits into a single commit before we merge

roles/contiv_network/tasks/ovs.yml
gaurav-dalvi, line 59: Would that break things for ubuntu ? Whole point of this PR : 
1: It should not break on Ubuntu (That @vvb found net_demo_installer script was failing because of this)
2: It should execute semanage command only when selinux is not disabled and when OS is redhat.

mapuri, line 59: >  It should not break on Ubuntu 

It won't due to `ignore_errors=true` and the conditional to check on the stdout, right?

> It should execute semanage command only when selinux is not disabled and when OS is redhat.

all this will be be ensured just by the check `selinux_set.stdout | match("SELinux status:.*enabled")`, right?

Actually thinking more, I think `ignore_errors=true` might not be the best way to do this as it will hide errors due to other command execution failures that we haven't envisioned. I think a better approach is following, wdyt?

```
- name: check selinux status
  shell: sestatus
  register: selinux_set
  when: (ansible_os_family == "RedHat")

- name: permit openvswitch_t type in selinux
  shell: >
       semanage permissive -a openvswitch_t
  when: (ansible_os_family == "RedHat") and (selinux_set.stdout | match("SELinux status:.*enabled"))
  become: true
```

mapuri, line 59: following fix will be needed in my approach above, if we go with that:

`match("SELinux status:.*enabled")` => `search(".*SELinux status:.*enabled.*")`

gaurav-dalvi: Looks good to me. But I need to hear from @vvb about ubuntu issue. I am pretty sure he ran in to that issue and even I was facing that issue so that was whole need to this PR.

@vvb : Could you please comment on following things.

1: on ubuntu, Are you able to execute net_demo_installer script correctly or not ?
2: On ReadHat / CentOs:
   2.1: Check net_demo_installer script when Selinux is disabled.
   2.2: Check net_demo_installer script when Selinux is enabled and is in enforcing mode
   2.3 : Check net_demo_installer script when Selinux is enabled and is in Permissive mode.

I dont have ubuntu setup. If you have please point me the details and I will be happy to test this.

mapuri:
thanks @gaurav-dalvi for making the changes.

LGTM

PR #188, opened 2016-05-09
Aci logging

gaurav-dalvi:
Adding -t option while starting aci-gw container. So that we can inspect logs of this container by doing docker logs -f aci-gw.

Right now because -t option is not present , whatever is getting printed at stdin in apiagent.py file, is not getting printed.


mapuri:
LGTM

PR #199, opened 2016-05-19
Nfs patches

erikh:
ignore the merge commits, they are harmless and it's github's fault.


mapuri:
@erikh I think we already have all the changes in the last commit. Also there are no files in the PR, which kind of indicates that too :)

erikh:
oh ahhahaha

PR #200, opened 2016-05-19
don't copy clusterm conf file if it exists

### LINKED ISSUE #198: on clusterm install/update the config file get's overwritten
this causes user's changes to be overwritten, it will be desirable to not overwrite the file if it exists

mapuri:
fixes #198

### DIFF
--- a/roles/contiv_cluster/tasks/main.yml
+++ b/roles/contiv_cluster/tasks/main.yml
@@ -45,12 +45,12 @@
 
 - name: copy conf files for clusterm
   copy:
-    src: "{{ item }}"
-    dest: /etc/default/clusterm/{{ item }}
-    force: yes
+    src: "{{ item.file }}"
+    dest: "/etc/default/clusterm/{{ item.file }}"
+    force: "{{ item.force }}"
   with_items:
-    - "{{ clusterm_args_file }}"
-    - "{{ clusterm_conf_file }}"
+    - { file: "{{ clusterm_args_file }}", force: "yes" }
+    - { file: "{{ clusterm_conf_file }}", force: "no" }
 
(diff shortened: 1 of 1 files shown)


vvb:
@mapuri the changes look good. I was just wondering about the case where a user really wants to overwrite the existing configuration. Should we be checking if the existing file is different from the supplied one. Or should we rely on user to delete the existing config manually before the new one gets applied. Worst case, should we just throw a warning/error if user supplied a config file but an config already pre-existed? Just so that the user does not assume that the new config has taken affect.

mapuri:
@vvb 

> I was just wondering about the case where a user really wants to overwrite the existing configuration

Yeah, I had thought about this a bit.

Right now, all we copy is an empty config which makes clusterm run with defaults (and this will most likely not change in future as well). In the most common deployment cases (i.e. baremetal deployments) a user _always_ needs to override this file. So it is almost always safe to keep the user's changes. 

Except when configuration options change in backward incompatible way (like it happened recently with `playbook_location`), in which case we would probably just be fine if we release note the fact. Note that even in this case overriding the file by default won't help and we would still need user to make the change themselves (or provide them a migration script, if our config becomes too complex).

> Should we be checking if the existing file is different from the supplied one. Or should we rely on user to delete the existing config manually before the new one gets applied. Worst case, should we just throw a warning/error if user supplied a config file but an config already pre-existed?

If a user wants to get rid of their config (in other unknown cases) I think we are better off just letting them do that themselves than us trying to do more checks, as we won't be able to cover all the cases anyways. So covering a 90% case (while having a way out for rest 10%) seems more compelling here as it keeps the handling simple.

vvb:
SGTM

PR #203, opened 2016-05-20
run etcd inside a container

mapuri:
there are no changes to the etcd setup logic other than it's dependency on docker now. Basic testing shows it works fine.

Note that we still download the corresponding etcd release binaries (along with container images) to get etcdctl.

I have also added a Makefile for common ansible test targets.

/cc @erikh @vvb

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+.PHONY: test-up test-provision test-cleanup
+
+test-up:
+	vagrant up
+
+test-provision:
+	vagrant provision
+
+test-cleanup:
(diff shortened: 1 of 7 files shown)


erikh:
Hey, sorry I never got to this.

Changes look good. I like the touch of installing etcdctl on the host, too.

mapuri:
thanks @erikh 

Sorry I hurried the merge a bit. This is our first core infra service to run in containers, so I wanted it tested in more setups sooner to weed out any unforeseen issues with the newly added docker dependency.

PR #209, opened 2016-05-26
temporarily use netctl for setting aci mode

mapuri:
this is to temporarily address part of https://github.com/contiv/netplugin/issues/388 caused by outdated contivctl while we work on the fix for https://github.com/contiv/contivctl/issues/3

/cc @erikh @rkharya @shaleman


jainvipin:
LGTM!

PR #214, opened 2016-06-02
Misc fixes and enhancements

mapuri:
- install a few base packages
  - e2fsprogs that provides mkfs.ext4 needed by docker
  - openssh-server provides sshd needed when provisioning a container. It is otherwise should already be installed.
- use 'become' in place of 'sudo', to cut on a few warnings
- use version variables for vbox, vagrant and packer

This is a preliminary patch that will enable us to use ansible to generate container images for CI, but I think it is a good to have patch in general.


erikh:
LGTM! 👍

Issue #201, opened 2016-05-19
UCP root CA keeys need to be copied to replicas


mapuri:
tracking this here based on recent email exchange with Uday from Docker team.

```
In UCP 1.1.0, the Root CA key is expected to be replicated across all replicas for HA, otherwise you tend to see the warning msg. Not sure if you've  added backup and restore in your scripts. 

docker run --rm --name ucp \ -v /var/run/docker.sock:/var/run/docker.sock \ docker/ucp backup --root-ca-only --passphrase "secret" --id $INSTANCEID > /tmp/backup.tar

On replica nodes, need to run the following:

docker run --rm -i --name ucp \ -v /var/run/docker.sock:/var/run/docker.sock \ docker/ucp restore --root-ca-only --passphrase "secret" --id $INSTANCEID < /tmp/backup.tar
```

vvb:
resolved by #216

PR #216, opened 2016-06-03
Adding support for backup and restore of ucp certificates

vvb:
PR for issue #201


roles/ucp/templates/ucp.j2
```diff
@@ -69,6 +69,14 @@ start)
     fi
     echo fingerprint: ${fingerprint}
     echo ${fingerprint} > "{{ ucp_remote_dir }}/{{ ucp_fingerprint_file }}"
+
+    # Save the root CA Key on bootstrap node
+    # The root CA key is replicated to replicas for HA
+    /usr/bin/docker run --rm -i --name ucp \
```
mapuri, line 75: hmm, I see an issue but it is not just limited to this PR I think, we can perhaps track it as separate issue.

We use `--fresh-install` flag for service start which will cause fingerptint, instance-id and certs to be regenerated (with different values) and might break the UCP running on other nodes. `--fresh-install` was added based on Uday's suggestion to overcome some UCP re-install issues, but may be we will need a better fix from docker folks here.

roles/ucp/tasks/cleanup.yml
```diff
@@ -9,6 +9,8 @@
   with_items:
     - "{{ ucp_fingerprint_file }}"
     - "{{ ucp_instance_id_file }}"
+    - "{{ ucp_certificate_file }}"
+    - "{{ ucp_fifo_file }}"
```
mapuri: this might be a merge thing but we don't need to cleanup `ucp_fifo_file` here as it is cleaned up when UCP service is stopped.

mapuri:
LGTM, other than one minor comment on cleanup of fifo file.

roles/ucp/tasks/cleanup.yml
vvb: @mapuri yes, this came in via merge.. will remove it.

vvb:
@mapuri resolved the comment..

mapuri:
thanks @vvb LGTM

PR #224, opened 2016-06-10
fix for ucp failure while importing certs

vvb:
This fixes the ucp failure caused by

```
Jun 10 00:12:59 ucsb-blade3 ucp.sh[54685]: cannot enable tty mode on non tty input
```


mapuri:
LGTM

PR #225, opened 2016-06-13
install and start services required for nfs

vvb:
- Installs the necessary packages for nfs for (redhat and debian)
- Start the necessary services for nfs (redhat and debian)
- `cleanup.yml` will stop the services that were started earlier.

### DIFF
--- a/roles/nfs/tasks/cleanup.yml
+++ b/roles/nfs/tasks/cleanup.yml
@@ -1,0 +1,14 @@
+---
+- name: stop nfs services (redhat)
+  service: "name={{item}} state=stopped"
+  with_items:
+    - rpcbind
+    - nfs-server
+    - rpc-statd
+    - nfs-idmapd
+
+- name: stop nfs services (debian)
+  service: "name={{item}} state=stopped"
+  with_items:
+    - nfs-kernel-server
+    - nfs-common
--- a/roles/nfs/tasks/server.yml
+++ b/roles/nfs/tasks/server.yml
@@ -1,0 +1,36 @@
+---
+- name: install nfs server (redhat)
+  yum: "update_cache=yes name={{ item }}"
+  with_items:
+    - nfs-utils
+    - libnfsidmap
+  when: ansible_os_family == "RedHat"
+
+- name: install nfs server (debian)
+  apt: "update_cache=yes name={{ item }}"
+  with_items:
+    - nfs-kernel-server
+    - rpcbind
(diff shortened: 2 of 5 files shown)


roles/nfs/tasks/cleanup.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: stop nfs services (redhat)
+  service: "name={{item}} state=stopped"
+  with_items:
+    - rpcbind
+    - nfs-server
+    - rpc-statd
+    - nfs-idmapd
+
+- name: stop nfs services (debain)
```
mapuri: debain => debian

mapuri:
LGTM

/cc @erikh as well for a quick look as I am not very familiar with nfs dependencies.

erikh:
is systemd not already starting rpcbind or something?

vvb:
@erikh a lot of blogs mentioned that systemd on redhat doesn't do that..

erikh:
I didn't experience that personally, but I'd rather be safe than sorry. Just curious.

LGTM

roles/nfs/tasks/server.yml
```diff
@@ -0,0 +1,36 @@
...
+
+- name: install nfs server (debian)
+  apt: "update_cache=yes name={{ item }}"
+  with_items:
+    - nfs-kernel-server
+    - rpcbind
+  when: ansible_os_family == "Debian"
+
+- name: enable nfs server services (redhat)
+  service: "name={{item}} enabled=yes"
```
mapuri, line 17: thinking more, we should perhaps not enable these services as it will run them automatically at reboot. This will cause incorrect behavior if the node was decommissioned and cleaned-up earlier. wdyt?

As of now we don't enable (non-base) services by default as that ensures when a node get's rebooted (i.e. disappears from cluster), it can be brought back into cluster by cluster-manager which will ensure pushing the configuration as per the current cluster state.

Also, may be slightly unrelated if there are multiple masters, is it safe to start nfs-server on all nodes?

erikh, line 17: if contiv storage and NFS are required, these services must be running.

erikh, line 17: NFS servers are referenced directly, there is no auto-discovery, so there is no issue with master/slaves.

mapuri, line 17: okay sounds good, we can let it be like this then.

PR #228, opened 2016-06-15
fix race condition in ucp certificate file transfer

vvb:


mapuri:
nice catch, LGTM!

rkharya:
indeed a good catch, Sir!

PR #229, opened 2016-06-24
resolving issue-92 in contiv/demo repo

gaurav-dalvi:
1: Need to add some delay before we set ACI mode. Have see this issue : https://github.com/contiv/demo/issues/92 number of time.

2: Netctl global set is the correct way to set it in ACI mode.

### DIFF
--- a/roles/contiv_network/tasks/aci_tasks.yml
+++ b/roles/contiv_network/tasks/aci_tasks.yml
@@ -16,8 +16,8 @@
   action: shell netctl tenant ls
   register: tenant_result
   until: tenant_result.stdout.find("default") != -1
-  retries: 5
-  delay: 5
+  retries: 9
+  delay: 10
 
 - name: set aci mode
   shell: netctl global set --fabric-mode aci
--- a/roles/contiv_network/tasks/aci_tasks.yml
+++ b/roles/contiv_network/tasks/aci_tasks.yml
@@ -12,5 +12,12 @@
 - name: start aci-gw container
   service: name=aci-gw state=started
 
+- name: Contiv | Wait for netmaster
+  action: shell netctl tenant ls
+  register: tenant_result
+  until: tenant_result.stdout.find("default") != -1
+  retries: 5
+  delay: 5
+
 - name: set aci mode
-  shell: contivctl network global set --fabric-mode aci
+  shell: netctl global set --fabric-mode aci
--- a/roles/contiv_network/tasks/aci_tasks.yml
+++ b/roles/contiv_network/tasks/aci_tasks.yml
@@ -12,7 +12,7 @@
 - name: start aci-gw container
   service: name=aci-gw state=started
 
-- name: Contiv | Wait for netmaster
+- name: wait for netmaster to be ready
   action: shell netctl tenant ls
   register: tenant_result
   until: tenant_result.stdout.find("default") != -1


roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -12,5 +12,12 @@
 - name: start aci-gw container
   service: name=aci-gw state=started
 
+- name: Contiv | Wait for netmaster
+  action: shell netctl tenant ls
+  register: tenant_result
+  until: tenant_result.stdout.find("default") != -1
+  retries: 5
+  delay: 5
```
jojimt: I have seen that 25 sec is not enough in some cases. Please use 9 retries with a 10 sec delay or 18 retries with a 5 sec delay. Note that it will break out of the loop as soon as it succeeds. So, there is no impact when it's successful.

gaurav-dalvi: Ok makes sense. On failed setup it was taking too much time so I thought reducing it !

jojimt:
LGTM

roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -12,5 +12,12 @@
 - name: start aci-gw container
   service: name=aci-gw state=started
 
+- name: Contiv | Wait for netmaster
```
mapuri: may be the task name should be `wait for netmaster to be ready` (without `Contiv |` prefix) to be consistent with rest of tasks?

roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -12,5 +12,12 @@
...
+  action: shell netctl tenant ls
```
mapuri, line 16: let's use the `shell` module instead, to keep it consistent with rest of playbook.

roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -12,5 +12,12 @@
...
+  register: tenant_result
+  until: tenant_result.stdout.find("default") != -1
+  retries: 9
+  delay: 10
+
 - name: set aci mode
-  shell: contivctl network global set --fabric-mode aci
+  shell: netctl global set --fabric-mode aci
```
mapuri, line 23: just curious, what is the issue with using contivctl?

roles/contiv_network/tasks/aci_tasks.yml
gaurav-dalvi: Makes sense. Will make that change.

roles/contiv_network/tasks/aci_tasks.yml
gaurav-dalvi, line 23: We use netctl command to set ACI mode. Dont know anybody is using contivctl command in testing / at customer side as well.

mapuri, line 23: hmm, okay. Yes, we have had similar discussions earlier and for sometime now at least the idea was to show off a unified cli tool to the user which was the reason to use `contivctl` in ansible. And you are right that none of the projects internally use this tool (may be it is just a matter of updating docs).

I am okay using netctl here, if everyone else is on same page. And I would rather want to not have to maintain contivctl if no one is using it (i.e. remove it's installation from ansible).

ping to @jainvipin @shaleman @erikh as well. Hopefully we can close it this time :)

erikh, line 23: Yeah let's ditch it.

mapuri:
LGTM

PR #232, opened 2016-06-28
bump up the vbox version

mapuri:
/cc @unclejack 

I will upgrade the lab nodes during quiet hours when there are no running jobs


mapuri:
merging it as it seems to be working fine

PR #236, opened 2016-07-04
force download of docker-compose when out of version

vvb:

without the `force` parameter, `get_url` will not download file if the `dest` is a file and it already exists. Generally `force==yes` is not advisable when the download is over the internet, but in this specific scenario we have a `when` that will protect us from cases where we should not be forcing download.


mapuri:
LGTM

PR #237, opened 2016-07-05
simplify etcd deploy scripts

mapuri:
- replaced `cluster-init` and `member add` scenarios with just one common handling, this removes the need for explicit `etcd_cluster_init` variable. This was already being tested in cluster-mgr.
- moved etcd-proxy setup logic to use `service_vip`, this removes the need for specifying etcd-master's name and address for configuring a worker nodes.

/cc @vvb @erikh


erikh:
LGTM, doesn't look like there'll be any portability issues either.

mapuri:
> doesn't look like there'll be any portability issues either.

yep, this shall not affect existing setups or projects.

It will help simplify cluster-mgr too which had to keep track of these variables internally :)

PR #238, opened 2016-07-06
adding aci-gw as a variable in ansible script

gaurav-dalvi:
adding aci-gw as a variable in ansible script.
For more details please see this : https://github.com/contiv/demo/issues/96

### DIFF
--- a/roles/contiv_network/templates/aci_gw.j2
+++ b/roles/contiv_network/templates/aci_gw.j2
@@ -19,7 +19,7 @@ start)
     -e "APIC_EPG_BRIDGE_DOMAIN={{ apic_epg_bridge_domain }}" \
     -e "APIC_CONTRACTS_UNRESTRICTED_MODE={{ apic_contracts_unrestricted_mode }}" \
     --name=contiv-aci-gw \
-    contiv/aci-gw
+    {{ aci_gw_version }}
     ;;
 
 stop)


roles/contiv_network/templates/aci_gw.j2
```diff
@@ -19,7 +19,7 @@ start)
     -e "APIC_EPG_BRIDGE_DOMAIN={{ apic_epg_bridge_domain }}" \
     -e "APIC_CONTRACTS_UNRESTRICTED_MODE={{ apic_contracts_unrestricted_mode }}" \
     --name=contiv-aci-gw \
-    contiv/aci-gw
+    {{ aci_gw_version }}
```
mapuri, line 22: shouldn't this be: `contiv/aci-gw:{{ aci_gw_version }}` ?

Also there should be default for the version I think.

gaurav-dalvi, line 22: 1: If someone wants to use his own custom image then just tag wont be sufficient. As we control contiv docker hub account, Unless we push his image he wont be able to try (We can do this technically by building container natively, but for someone who is not from tech background it will be difficult.)

2: Default version is mentioned in net_demo_installer script.
https://github.com/contiv/demo/pull/97

mapuri, line 22: > If someone wants to use his own custom image then just tag wont be sufficient.

hmm, I am not sure if I understand. You mean someone can use a non aci-gw image built by them?

> Default version is mentioned in net_demo_installer script.

what I meant was ansible shall have a default defined too, as this will be use by other projects as well. Please se how it is done for other plays. For instance checkout etcd container download task: https://github.com/contiv/ansible/blob/master/roles/etcd/tasks/main.yml#L20

gaurav-dalvi, line 22: Is there any way in ansible so that i can have default version as : contiv/aci-gw and if user passes different version then it should overwrite and take this new version ? 

This will avoid this logic spreading at two different places.

mapuri, line 22: yes, variables are the way to do this and you set their defaults in the defaults dir. For `contiv_network` it will be in https://github.com/contiv/ansible/blob/master/roles/contiv_network/defaults/main.yml

Also if you want the user to specify a container image and not just version, please consider renaming `aci_gw_version` to `aci_gw_image`, though I feel we should keep image name and version separate.

gaurav-dalvi, line 22: one of the marketing person just used his own image as he was using different ACI version.
we had to log in to his setup and make sure that he fetches container :  roschaer/aci-gw:2.0_1m.

mapuri, line 22: > one of the marketing person just used his own image as he was using different ACI version. we had to log in to his setup and make sure that he fetches container

that is exactly my point. By allowing folks to specify an image this is bound to happen and we just increase the chances of failures.

Anyways if you would still like to give this flexibility you can use two variables instead, something like `{{ aci_gw_image }}: {{ aci_gw_version }}` and set reasonable defaults for both in the defaults file that I mentioned above

PR #239, opened 2016-07-07
adding aci-gw as a variable in ansible script

gaurav-dalvi:
Please refer issue : https://github.com/contiv/demo/issues/96 for more details

### DIFF
--- a/roles/contiv_network/defaults/main.yml
+++ b/roles/contiv_network/defaults/main.yml
@@ -16,6 +16,8 @@ bgp_port: 179
 vxlan_port: 4789
 netplugin_rule_comment: "contiv network traffic"
 
+aci_gw_version: "contiv/aci-gw"
+aci_gw_image: "latest"
 contiv_network_version: "v0.1-06-30-2016.20-09-48.UTC"
 contiv_network_tar_file: "netplugin-{{ contiv_network_version }}.tar.bz2"
 contiv_network_src_file: "https://github.com/contiv/netplugin/releases/download/{{ contiv_network_version }}/{{ contiv_network_tar_file }}"
--- a/roles/contiv_network/defaults/main.yml
+++ b/roles/contiv_network/defaults/main.yml
@@ -16,8 +16,6 @@ bgp_port: 179
 vxlan_port: 4789
 netplugin_rule_comment: "contiv network traffic"
 
-aci_gw_version: "contiv/aci-gw"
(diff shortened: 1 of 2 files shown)


roles/contiv_network/defaults/main.yml
```diff
@@ -16,6 +16,8 @@ bgp_port: 179
 vxlan_port: 4789
 netplugin_rule_comment: "contiv network traffic"
 
+aci_gw_version: "contiv/aci-gw"
+aci_gw_image: "latest"
```
mapuri: @gaurav-dalvi you probably meant to have these default values assigned the other way?

gaurav-dalvi: @mapuri : Already added via this PR : https://github.com/contiv/demo/pull/97

But i think we need this default in ansible. aci-gw container is very much specific to ACI + Contiv and I dont think it will change frequently (Tightly depends on ACI releases). Also in net_demo_installer script I have a way to change these default variables via shell variables (just the way we did it for contiv_network_version). 

So If someone does not specifiy, it will take contiv/aci-gw:latest and if someone wants to change he can set these same shell variables and net_demo_installer script will take it and overwrite these defaults here.

roles/contiv_network/templates/aci_gw.j2
```diff
@@ -19,7 +19,7 @@ start)
     -e "APIC_EPG_BRIDGE_DOMAIN={{ apic_epg_bridge_domain }}" \
     -e "APIC_CONTRACTS_UNRESTRICTED_MODE={{ apic_contracts_unrestricted_mode }}" \
     --name=contiv-aci-gw \
-    contiv/aci-gw
+    {{ aci_gw_image }}:{{ aci_gw_version }}
```
jojimt: Can we take this as a single variable instead? The whole thing is a string anyway.

gaurav-dalvi: yes. most of the time we will be changing the tags and rarely will be changing the docker image name. That was the point having different variables here. Let me know.

jojimt: I guess it's fine either way.

gaurav-dalvi: I think having one variable eliminates need to having tag=latest to be entered explicitly. 
let me repost PR. !

PR #241, opened 2016-07-09
adding aci_gw_image variable to ansible scripts

gaurav-dalvi:


jojimt:
Please add a default value: contiv/aci-gw:latest
Other than that, LGTM

gaurav-dalvi:
@jojimt : Default value is attached via this : 
https://github.com/contiv/demo/pull/98/files#diff-6a689c604ce54a47aab15671bd9c13acR152

jojimt:
We still need a default in ansible (otherwise, ansible can only be run from net_demo_installer).

gaurav-dalvi:
That wont work. I tested it. This is exactly similar change Reshma did it for contiv_network_version.
In that change we specified default value in net_demo_installer script and then we can overwrite by exporting shell variable. We dont have default value in ansible script though.

Reference : https://github.com/contiv/demo/pull/94

jojimt:
@gaurav-dalvi the difference is that it is not possible to have a default value for contiv network version in ansible(there is not latest tag). In this case, it is possible to have a default and I think it is better to have it.

gaurav-dalvi:
@jojimt : Updated the PR.

PR #243, opened 2016-07-12
Glusterfs

erikh:
This implements glusterfs support. It is a first draft and you should expect more changes to come in the future as I start doing more with it.

This is a very fast playbook because gluster is ridiculously easy to install. :D


mapuri:
thanks @erikh LGTM

Just listing a few items. I think they can all be tackled in future as need arises:
- based on similar discussion elsewhere in ansible repo (around docker device etc) I think may be configuration of brick mount points etc will need to come from user. 
- also perhaps we could move peer computation and probing logic out to it's own file and re-think incremental addition of peers to the cluster. But I think right now it is okay to leave it as is and do the changes as we see use cases.

erikh:
Ok, let's discuss this when the time is right. FWIW, to get snapshotting on
gluster we need to do the LVM thin provisioning we're doing for docker too.

PR #247, opened 2016-07-18
fixes missing iptables entries for serf on reboot

vvb:
The Iptables entries for serf are missing on a system reboot. This happens because we apply the rules too soon. This PR will make `iptables` wait until `network` service is up.

```
# iptables rule do not get applied on a reboot - probably because we try and apply them too early
[stack@contiv-b3 ~]$ cat /etc/systemd/system/serf.service | grep After
After=auditd.service systemd-user-sessions.service time-sync.target
[stack@contiv-b3 ~]$
[stack@contiv-b3 ~]$ sudo reboot

# post reboot, the iptables are missing
[stack@contiv-b3 ~]$ sudo iptables -L | grep serf
[stack@contiv-b3 ~]$



# adding a After=network.target sovles this issue
# given that they are needed only after network is up.
[stack@contiv-b3 ~]$ cat /etc/systemd/system/serf.service | grep After
After=network.target auditd.service systemd-user-sessions.service time-sync.target
[stack@contiv-b3 ~]$

# reboot
[stack@contiv-b3 ~]$ sudo reboot

# Login and check iptables
[stack@contiv-b3 ~]$ sudo iptables -L | grep serf
ACCEPT     tcp  --  anywhere             anywhere             tcp dpt:7946 /* serf control */
ACCEPT     udp  --  anywhere             anywhere             udp dpt:mdns /* serf discovery dport */
ACCEPT     udp  --  anywhere             anywhere             udp spt:mdns /* serf discovery sport */
[stack@contiv-b3 ~]$
```

### DIFF
--- a/roles/serf/files/serf.service
+++ b/roles/serf/files/serf.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Serf
-After=auditd.service systemd-user-sessions.service time-sync.target
+After=network.target auditd.service systemd-user-sessions.service time-sync.target
 
 [Service]
 ExecStart=/usr/bin/serf.sh start


mapuri:
thanks @vvb 

LGTM

roles/serf/files/serf.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Serf
-After=auditd.service systemd-user-sessions.service time-sync.target
+After=network.target auditd.service systemd-user-sessions.service time-sync.target
```
mapuri, line 3: @vvb while this change is good, 

I was thinking why didn't the script fail in first place, which would have caused systemd to retry the service as well. It turns out there is a bug in the statement that install iptables rule:

```
( /sbin/iptables -L INPUT | grep {{ mdns_sport_comment }} || \
/sbin/iptables -I INPUT 1 {{ mdns_sport_rule }} )
```

should have been:

``` bash
( /sbin/iptables -L INPUT | grep {{ mdns_sport_comment }} || \
  ( /sbin/iptables -I INPUT 1 {{ mdns_sport_rule }} || exit 1) )
```

we can fix this separately or in this PR, upto you.

vvb, line 3: @mapuri makes sense.. Let me try it out standalone first, without `network.target` dependency to see if it works the way it should. and will do both the changes in this PR itself.

vvb, line 3: @mapuri I tried with the below, it does not seem to hit the exit. so, no iptables entries installed post reboot

```
    echo setting up iptables for serf
    ( /sbin/iptables -L INPUT | grep 'serf discovery sport' || \
        ( /sbin/iptables -I INPUT 1 -p udp --sport 5353 -i enp133s0 -j ACCEPT -m comment --comment 'serf discovery sport' || exit 1) )
    ( /sbin/iptables -L INPUT | grep 'serf discovery dport' || \
        ( /sbin/iptables -I INPUT 1 -p udp --dport 5353 -i enp133s0 -j ACCEPT -m comment --comment 'serf discovery dport' || exit 1) )
    ( /sbin/iptables -L INPUT | grep 'serf control' || \
        ( /sbin/iptables -I INPUT 1 -p tcp --dport 7946 -i enp133s0 -j ACCEPT -m comment --comment 'serf control' || exit 1) )
```

mapuri, line 3: hmm, that's a bummer. Thanks for checking @vvb 

May be this command doesn't return with non-zero code(I am not able to validate this as I can't seem to disable iptables in a running system) or may be when firewalld comes up it wipes out existing rules (this seems more likely). 

I think what you have works then.

mapuri:
LGTM, will merge this.

PR #249, opened 2016-07-20
Ubuntu1604 (Xenial) support

mapuri:
This PR brings in the changes required to support Ubuntu 1604 in ansible.

The change is mostly in Vagrantfile to enable the CI with Ubuntu 15.10 and 16.04 flavors.

There are no major functional changes except for:
- handling docker CS engine repos as Docker CS engine support for Xenial seems to be only from version 1.11 onwards.
- moving environment file setup for docker daemon ahead of  installation task. This ensures that docker service get's started with updated environment.

/cc @contiv/core-storage @contiv/core-network @vvb

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -23,42 +23,57 @@ ansible_tags =  ENV["CONTIV_ANSIBLE_TAGS"] || "prebake-for-dev"
 ansible_extra_vars = {
     "env" => host_env,
     "validate_certs" => "no",
+    "control_interface" => "eth1",
 }
 
 puts "Host environment: #{host_env}"
 
 Vagrant.configure(2) do |config|
-    node_name = "host0"
-    config.vm.define node_name do |node|
-        node.vm.hostname = node_name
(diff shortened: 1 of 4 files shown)


erikh:
LGTM, wish we could find a solution that didn't require removing packer from a third-party image, though.

mapuri:
@erikh 

I agree. The problem is that there seems to be a another binary with name `packer` in certain centos images that causes issues while installing `packer` that we need :) (https://github.com/mitchellh/packer/issues/1117#issuecomment-42138655)

Actually, I just realized that we don't install packer as part of ansible CI now (as the packer installation has been moved to `test` role, that we only use internally). So, I have updated the patch to remove that work-around.

erikh:
Yeah, I'd double check to ensure `packer` isn't a different system-level binary and not hashicorp's tool.

mapuri:
agree, let me file an issue to handle this in `packer` installation task.

vvb:
LGTM

PR #252, opened 2016-07-25
Global forwarding mode knob 

abhi:
Adding global setting to change between routing vs bridging modes

### DIFF
--- a/roles/contiv_network/tasks/main.yml
+++ b/roles/contiv_network/tasks/main.yml
@@ -108,3 +108,8 @@
 
 - include: aci_tasks.yml
   when: (run_as == "master") and (contiv_network_mode == "aci")
+
+- name: set fwd mode
+  shell: netctl global set --fwd-mode "{{ fwd_mode }}"
+  when: (run_as == "master")
+  run_once: true 
--- a/roles/contiv_network/defaults/main.yml
+++ b/roles/contiv_network/defaults/main.yml
@@ -17,7 +17,7 @@ vxlan_port: 4789
 netplugin_rule_comment: "contiv network traffic"
 
 aci_gw_image: "contiv/aci-gw"
-contiv_network_version: "v0.1-07-14-2016.07-06-17.UTC"
+contiv_network_version: "v0.1-08-06-2016.07-13-09.UTC"
 contiv_network_tar_file: "netplugin-{{ contiv_network_version }}.tar.bz2"
 contiv_network_src_file: "https://github.com/contiv/netplugin/releases/download/{{ contiv_network_version }}/{{ contiv_network_tar_file }}"
 contiv_network_dest_file: "/tmp/{{ contiv_network_tar_file }}"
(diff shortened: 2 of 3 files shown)


roles/contiv_network/tasks/main.yml
```diff
@@ -108,3 +108,6 @@
 
 - include: aci_tasks.yml
   when: (run_as == "master") and (contiv_network_mode == "aci")
+
+- include: fwd_tasks.yml
+  when: (run_as == "master")
```
mapuri: I am not sure if we want this task to be conditional. This condition will make the task run only on `master` nodes (and we run netplugin on `worker` nodes a well). Since the flag is a netplugin flag, I think you want to run it on all nodes, right?

roles/contiv_network/tasks/fwd_tasks.yml
```diff
@@ -0,0 +1,3 @@
+- name: set fwd mode
+  shell: netctl global set --fwd-mode "{{ fwd_mode }}" 
+  run_once: true
```
mapuri: @abhinandanpb `run_once` will cause the task to be run only on one node among all nodes that are present in a host-group at the time ansible is run. Is that what you want?

roles/contiv_network/tasks/fwd_tasks.yml
```diff
@@ -0,0 +1,3 @@
+- name: set fwd mode
```
mapuri: I think we shall just have this task in `main.yml` than putting in a separate file. wdyt?

roles/contiv_network/tasks/main.yml
abhi: no this is a global setting. Doesnt need to be run on all nodes. It just needs to be run on master.

roles/contiv_network/tasks/fwd_tasks.yml
abhi: @mapuri yep. It just needs to be run on master. So we should be good.

roles/contiv_network/tasks/fwd_tasks.yml
abhi: sure. I was thinking the convention was to add it in seperate files similar to aci. and eventually if we end up passing some args and stuff to the global setting it might be cleaner. your call.

roles/contiv_network/tasks/fwd_tasks.yml
mapuri: ah, got it.

sorry, I missed that.

roles/contiv_network/tasks/fwd_tasks.yml
mapuri: yeah, as of now we have been keeping the common tasks in `main.yml` while deployment specific tasks (like the ones for aci) in respective files. Since this seems like common task I was thinking of leaving it in `main.yml`. Are  you okay with making this change?

mapuri:
@abhinandanpb LGTM other than one minor comment on moving the task to main.yml

Also can you squash your changes into a single commit? That will keep the commit history clean.

roles/contiv_network/tasks/main.yml
```diff
@@ -108,3 +108,8 @@
 
 - include: aci_tasks.yml
   when: (run_as == "master") and (contiv_network_mode == "aci")
+
+- name: set fwd mode
+  when: (run_as == "master")
```
mapuri: sorry to nit-pic, suggest to move `when` condition after `shell:`, to look consistent with rest of tasks.

mapuri:
@abhinandanpb do we need to bump up the default version of contiv_network [here](https://github.com/contiv/ansible/blob/master/roles/contiv_network/defaults/main.yml#L20) to be able to use this knob?

abhi:
@mapuri I will update the contiv version after the version of contiv binary having this change is ready. Will do it all together.

abhi:
@mapuri i have updated to the lastest binary and the suggested changes.

mapuri:
LGTM

PR #255, opened 2016-07-26
Consul

erikh:
This is initial support for running consul on small clusters.

It is most certainly not intended to be a long-term solution. I will leave that to the experts. :D 

Please let me know if there's anything you'd like me to do with this patch.


erikh:
cc @jainvipin

mapuri:
thanks @erikh 

We need to add cleanup.yml that can take care care of stopping the service and cleaning up ip table rules. Otherwise LGTM !

erikh:
@mapuri PTAL

mapuri:
LGTM, thanks @erikh

PR #260, opened 2016-08-10
don't use epel-testing repo

### LINKED ISSUE #259: install ansible from epel-release instead of epel-testing
epel-testing now has ansible version 2.1.1, which leads to the below error

```
FAILED! => {\"failed\": true, \"reason\": \"'item' is undefined\"}"
```

we need to stick to ansible 2.1.0 which is not available in `epel-release`

mapuri:
fixes #259


gaurav-dalvi:
LGTM

PR #262, opened 2016-08-11
add support for kubernetes scheduler stack

### LINKED ISSUE #9: need ways to introduce k8s
The roles and groups that are defined so far includes swarm scheduler. What would be the best way to contribute to this repo to add k8s scheduler. Few options I can think of:
- create host-groups that are different for k8s, for example host-group service-master, can be split into two 'service-master-docker' vs 'service-master-k8s'
- abstract out scheduler role from this to be called 'scheduler' and replace it with swarm, k8s, etc.

My preference would be first one (different host-groups) because there may be more dependent top level roles that some of the schedulers (e.g. mesos) might require.

ping @mapuri, @erikh, @jojimt

mapuri:
**Changes**:
- run kubernetes in container using hyperkube (https://github.com/kubernetes/kubernetes/tree/master/cluster/images/hyperkube)
  - the master and worker services are started as pod specs/manifests
    - master services include apiserver, controller-manager and scheduler
    - worker services include kube-proxy and kubelet
    - apiserver is run behind a VIP for HA
    - scheduler and controller-manager use `--leader-elect` flag to select the active leader
    - kube-proxy and kubelet is run on every node
  - default certs are packaged but allows user to override with their own by setting `k8s_local_cert_dir` var
- contiv_network changes to set the scheduler mode and setup k8s specific config
  - also moved the task to wait on netmaster ready to common tasks, otherwise global mode setting fails occasionally (this is more of a independent fix, so I have left it as separate commit)
- docker service is started with `MountFlags=shared` to propagate mount events to Kubernetes containers

**Testing**:
I tried creating a few pods with contiv networking and basic ping tests passed!

**Next steps**:
Once we start using this more we might find more things to tweak or add (like I haven't enabled k8s cluster-dns (for service discovery) or add-ons yet). We can address those in subsequent PRs.

fixes #9 

/cc @jainvipin @erikh @vvb


erikh:
wow. nice. I'll review tonight.

mapuri:
folks, let me know if there are any major comments. Otherwise I think we can merge this and start using it in our projects to identify next set of issues/enhancements.

erikh:
sorry looking at this now

roles/kubernetes/files/certs/ca.crt
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
erikh, line 1: are these overridable?

roles/kubernetes/tasks/cleanup.yml
```diff
@@ -0,0 +1,11 @@
+---
+# This play contains tasks for cleaning up kubernetes
+
+- name: stop kubernetes
+  service: name=kubernetes state=stopped
+
+- name: cleanup iptables for kubernetes
```
erikh, line 7: we should really turn this into a single unit of code in the future; this pattern is all over the place in our code now :(

roles/docker/templates/docker-svc.j2
```diff
@@ -12,7 +12,7 @@ ExecStart=/usr/bin/docker daemon --storage-opt dm.thinpooldev=/dev/mapper/contiv
 {% else %}
 ExecStart=/usr/bin/docker daemon -H fd:// --cluster-store=etcd://localhost:{{ etcd_client_port1 }}
 {% endif %}
-MountFlags=slave
+MountFlags=shared
```
vvb: @mapuri shall we make this an ansible variable with defaults to `shared`? for cases where people are not using k8s or volplugin and where using `slave` should be ok. wdyt?

vvb:
@mapuri LGTM. really good work.

erikh:
LGTM minus my queries. Very nice.

roles/kubernetes/files/certs/ca.crt
mapuri, line 1: yep, all the certs and keys can be overridden by setting `k8s_local_cert_dir` variable. The bundled certs will be helpful for bringing up test and demo environments

roles/kubernetes/tasks/cleanup.yml
mapuri, line 7: yes, I agree. Let me see if I can make use of a common template of some sort.

Let me track this separately so I can fix all roles together.

roles/docker/templates/docker-svc.j2
mapuri: @vvb 

Yes, both volplugin and k8s need this mode. I didn't really see a good usecase for `slave` during my search so I left it hardcoded. Let me make it a variable.

erikh: slave is technically safer.

erikh: I should clarify that -- we _need_ shared, but for those who don't (e.g., netplugin-only consumers on swarm) probably don't want it.

mapuri:
@vvb I have made the `MountFlags` come through a variable.

I will merge this now and start working on integrating it into cluster-mgr. 

/cc @abhinandanpb @gaurav-dalvi : you might be interested in incorporating this into the demo installer. It can help consolidate some of the demo and test scripts. Let me know if I can help in any way.

vvb:
@mapuri thanks for taking care of it.

PR #267, opened 2016-08-18
Misc fixes and enhancements

### LINKED ISSUE #264: epel-release repo is also installing ansible 2.1.1.0 now
Reference issue : https://github.com/contiv/ansible/issues/259

epel-testing to epel-release repo transition wont work now.  ansible 2.1.1.1 version gets installed and then our ansible scripts will fail because of this error :

```
TASK [include] *****************************************************************
included: /home/admin/ansible/roles/ucarp/tasks/cleanup.yml for node1, node2
fatal: [node1]: FAILED! => {"failed": true, "reason": "'item' is undefined"}
...ignoring
fatal: [node2]: FAILED! => {"failed": true, "reason": "'item' is undefined"}
```

mapuri:
- fix the cleanup tasks by expanding the loop of includes. Looks like this is might be broken in latest  ansible, so filed an issue there as well: https://github.com/ansible/ansible/issues/17144
- cleaned up some extra ignore_errors fields, that are not needed anymore with ansible 2.0 +

fixes #264

### DIFF
--- a/cleanup.yml
+++ b/cleanup.yml
@@ -15,15 +15,34 @@
         - "ucp"
         - "docker"
         - "etcd"
-    - include: roles/{{ item }}/tasks/cleanup.yml
-      with_items:
-        - ucarp
-        - contiv_network
-        - contiv_storage
-        - swarm
-        - kubernetes
-        - ucp
-        - etcd
-        - nfs
-        - docker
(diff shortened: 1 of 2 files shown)


vvb:
LGTM

PR #271, opened 2016-09-05
Bind

erikh:


mapuri:
LGTM

@erikh In the volplugin repo there was some discussion on using a different tool but it isn't closed yet. We can merge this whenever you are ready.

dseevr:
I think dnsmasq is a better choice in general (far fewer CVEs... lol), but for what this is being used for, it doesn't really matter IMO

erikh:
I agree with bill on literally everything here

this is dependent on @mapuri's merge into contiv/ansible which is why I'm waiting.

erikh:
oh! this is contiv/ansible. ok. merging.

PR #275, opened 2016-11-04
aws.tf,.gitignore: Terraform AWS environment to run netplugin!

erikh:
This supplies an `aws.tf` at hte root of the tree that provisions the ansible to N hosts (3 by default) in AWS.

It can be run by configuring your AWS shell environment, checking out this branch, installing [terraform](terraform.io) and typing:

```
$ terraform apply .
```

When you are ready to destroy it:

```
$ terraform destroy .
```

To get the IP addresses of the hosts (and a lot of other info):

```
$ terraform show
```

**the terraform.tfstate file is very important to keep! do NOT delete it!**

This will provision all the resources on AWS it needs and open ssh to the world with the key information you provide when asked (you can also supply this on the commandline if you find it annoying). It will then run an ansible plan against the hosts to install netmaster on the first host and netplugin's stack on all of them.

This isn't as pretty as I'd like, but it's working well in practice so far. I'm not super sure if the vlan support actually works on AWS, I didn't test. However, overlay networking operates as expected.

The ansible was modified to remove many volplugin requirements that lived in the netplugin namespace, these can be re-added at a later time if desired.

Please try this branch before reviewing it, unless you are very comfortable speaking terraform. It's a complicated system that's hard to understand at first.

### DIFF
--- a/roles/docker/defaults/main.yml
+++ b/roles/docker/defaults/main.yml
@@ -11,4 +11,5 @@ docker_device_metadata_size: "1000MB"
 docker_cs_engine_supported_versions_ubuntu:
     "wily": [ "1.10", "1.11" ]
     "xenial": [ "1.11" ]
+    "Core": [""] # needed for redhat installations due to templating
 docker_mount_flags: "shared"
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ fetch/*
 
 #vagrant temp files
 .vagrant*
+
+terraform.tfstate*
--- a/aws.tf
+++ b/aws.tf
@@ -1,0 +1,180 @@
+provider "aws" {
+  region = "${var.aws_region}"
+}
+
+variable "nodes" {
+  default = 3
+}
+
+variable "ami" {
+  default = "ami-af4333cf"
+}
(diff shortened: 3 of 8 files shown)


erikh:
I lost a commit or something here so I'm working that out now.

erikh:
made some changes here; basically, ucarp installation was bugging the interfaces out on AWS, so I fixed it by providing a direct netmaster ip which if not set, consults the service_vip (which must be supplied) and installs ucarp. Otherwise it does neither.

storage was also pruned since it is not necessary atm.

roles/docker/defaults/main.yml
```diff
@@ -11,4 +11,5 @@ docker_device_metadata_size: "1000MB"
 docker_cs_engine_supported_versions_ubuntu:
     "wily": [ "1.10", "1.11" ]
     "xenial": [ "1.11" ]
+    "Core": [""] # needed for redhat installations due to templating
```
gaurav-dalvi, line 14: is it new ansible version specific thing ?

erikh: yes

PR #288, opened 2016-12-25
Update main.yml

gaurav-dalvi:


dvavili (commented):
How are we handling the upgrade scenarios? Do the service files get rewritten? Otherwise we might run into issues when the swarm container is started again.

gaurav-dalvi:
@DivyaVavili : good point. 
We are not doing upgrade here.  Its changing current swarm version to 1.2.5 for all future setups. There is not version later than 1.2.5.

gaurav-dalvi:
I have tested this. After installation is complete, u see swarm 1.2.5 container is up.

dvavili:
@gaurav-dalvi: The scenario I was mentioning is if you ran a cleanup task and then try to re-run the ansible task. In that case, it would be similar to a upgrade scenario from the previous version coz the cleanup tasks don't remove the systemd files.

Anyway, I checked the default behavior for the copy and template modules that we use and we seem to be overwriting files if anything has changed. So, we should be good to go and the changes LGTM.

PR #290, opened 2017-01-06
add dynamic loading to include modules

dvavili:
Add dynamic loading of include modules. The behavior of include modules changed in Ansible 2.1.
Include ignore_errors for ovs_cleanup tasks.

### DIFF
--- a/cleanup.yml
+++ b/cleanup.yml
@@ -15,34 +15,16 @@
         - "ucp"
         - "docker"
         - "etcd"
-    - include: roles/ucarp/tasks/cleanup.yml
+    - include: roles/{{ item }}/tasks/cleanup.yml
+      with_items:
+        - ucarp
+        - contiv_network
+        - contiv_storage
+        - swarm
+        - kubernetes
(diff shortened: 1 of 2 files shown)


gaurav-dalvi:
Thanks for taking care of this @DivyaVavili .
Do we need to do this change in all of the ansible repo, wherever we are using ignore_error ?
(all cleanup.ymls for each role ? )

dvavili:
It's needed only for include modules. The ignore_errors directive in tasks will continue to function as before.

gaurav-dalvi:
sorry I meant include modules. Do we need to modify at all places in this repo ?

dvavili:
Lets not make this change everywhere. Dynamic loading seems to have different effects. If we hit other issues in which we see that the variables/directives are not getting inherited by included tasks, we will know what to fix.

unclejack:
I've tested this and it seems to be OK.
LGTM

gaurav-dalvi:
Tested this and it works !

PR #293, opened 2017-01-06
Adding golang role to ansible

gaurav-dalvi:
Signed-off-by: Gaurav Dalvi <gaurav1424@gmail.com>

### DIFF
--- a/roles/golang/defaults/main.yml
+++ b/roles/golang/defaults/main.yml
@@ -1,0 +1,9 @@
+---
+
+# Role defaults for golang. Installation of specific version of golang
+# also things needed to compile netplugin code
+# setting up GOPATH and PATH variables
+
+golang_version: "1.7.4"
+golang_download_url: "https://storage.googleapis.com/golang/go{{ golang_version }}.linux-amd64.tar.gz"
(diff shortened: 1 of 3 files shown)


unclejack:
This can be found here https://github.com/contiv/ansible/blob/master/roles/dev/tasks/os_agnostic_tasks.yml already.

I don't agree, nor like to duplicate this code in two places. Why is this needed? We should use the existing code if it's needed for other roles than dev.

gaurav-dalvi:
@unclejack : Good find. We need to have golang setup on one of the machines in our setup. We are trying to replicate our current vagrant approach. 

Someone send PR -- > Will trigger 3 VMs (e.g: swarm + contiv) then --> will compile code on node 1 and built binary and then start system-tests on that setup.

Current vagrant box has inbuilt golang installed in it.

gaurav-dalvi:
I will try to hook this role into our installation yml.

unclejack:
@gaurav-dalvi It's ok, I can make this change if you need it. I've got some other changes batched for ansible.

unclejack:
I'm closing this in favor of issue #295.

PR #298, opened 2017-01-10
Support RHEL 7.x

jojimt:
Changes to make sure install works on RHEL as well.

### DIFF
--- a/roles/docker/tasks/redhat_install_tasks.yml
+++ b/roles/docker/tasks/redhat_install_tasks.yml
@@ -9,11 +9,8 @@
     validate_certs: "{{ validate_certs }}"
 
 - name: add docker CS-engine repos (redhat)
-  shell: yum-config-manager --add-repo https://packages.docker.com/{{ item }}/yum/repo/main/centos/7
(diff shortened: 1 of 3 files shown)


gaurav-dalvi:
how are we handling (if at all) licensing of RHEL ? considering we are going to create RHEL VMs for testing.

jojimt:
We are definitely not providing licenses :)

roles/docker/tasks/redhat_install_tasks.yml
```diff
@@ -9,11 +9,8 @@
     validate_certs: "{{ validate_certs }}"
 
 - name: add docker CS-engine repos (redhat)
-  shell: yum-config-manager --add-repo https://packages.docker.com/{{ item }}/yum/repo/main/centos/7
+  shell: yum-config-manager --add-repo https://yum.dockerproject.org/repo/main/centos/7/
```
gaurav-dalvi, line 12: I hope this does not break current centos 7 provisioning !

jojimt: I have tested with centos and rhel.

PR #303, opened 2017-01-13
Add status logs for etcd & netmaster to aid debugging

neelimamukiri:
We are frequently seeing installs fail with netmaster not starting.
Adding logs to make debugging such cases simpler.

### DIFF
--- a/roles/contiv_network/tasks/main.yml
+++ b/roles/contiv_network/tasks/main.yml
@@ -107,6 +107,24 @@
   retries: 9
   delay: 10
 
+# get the netmaster status for debugging in case there are errors
+- name: get netmaster status
+  shell: systemctl status netmaster -l
+  register: netmaster_status
+  when: (run_as == "master")
+
+- debug: msg={{ netmaster_status.stdout }} verbosity=2
+  when: (run_as == "master")
+  ignore_errors: true
+
+# get the netplugin status for debugging in case there are errors
+- name: get netplugin status
+  shell: systemctl status netplugin -l
+  register: netplugin_status
+
+- debug: msg={{ netplugin_status.stdout }} verbosity=2
+  ignore_errors: true
+
 - include: aci_tasks.yml
   when: (run_as == "master") and (contiv_network_mode == "aci")
 
--- a/roles/etcd/tasks/main.yml
+++ b/roles/etcd/tasks/main.yml
@@ -41,4 +41,13 @@
 - name: start etcd
   service: name=etcd state=started
 
+- name: wait for etcd to startup
+  shell: systemctl status etcd -l
+  register: etcd_result
+  until: etcd_result.stdout.find("Active") != -1
+  retries: 9
+  delay: 10
+
+- debug: msg={{ etcd_result.stdout }} verbosity=2
+  ignore_errors: true
 


roles/contiv_network/tasks/main.yml
```diff
@@ -107,6 +107,10 @@
   retries: 9
   delay: 10
 
+# get the netmaster status for debugging in case there are errors
+- name: get netmaster status
+  shell: systemctl status netmaster -l
+
```
gaurav-dalvi: How does it different than !

https://github.com/contiv/ansible/blob/master/roles/contiv_network/tasks/main.yml#L103

gaurav-dalvi:
This is good. Dont we want to do the same for netplugin service as well ?

neelimamukiri:
I will add netplugin as well.

roles/contiv_network/tasks/main.yml
```diff
@@ -107,6 +107,15 @@
   retries: 9
   delay: 10
 
+# get the netmaster status for debugging in case there are errors
+- name: get netmaster status
```
gaurav-dalvi, line 111: how does it different from https://github.com/contiv/ansible/blob/master/roles/contiv_network/tasks/main.yml#L102

unclejack: +1
Can this be reused instead?

neelimamukiri: @gaurav-dalvi, @unclejack, Sorry I missed this update. "shell netctl tenant ls" makes sure that netmaster is up and there is a default tenant. "systemctl status netmaster -l" gets the systemd service status and the logs for the service. In some cases when netmaster does not come up, the service may be down or sometimes the logs show what went wrong even if the service is active. So they serve different purposes, and this status is only for logging/debugging.

unclejack: Thank you for the explanation, @neelimamukiri.

unclejack:
@neelimamukiri: The setup for the demo repository is failing every single time at the end with these changes. The last task (contiv-networking, from what I recall) fails on all nodes every single time. I've tried it three times and it failed every single time.

Can you take a look when you have the time to investigate these failures, please?

neelimamukiri:
@unclejack, I am unable to repro the issue. Looking at the logs you sent(...template error while templating string: expected token 'end of print statement', got ':'. String: ● netplugin.service - ...), looks like we ran into an Ansible parsing issue.
https://github.com/ansible/ansible/blob/9abdb1bb7c1417d05f357de22c99b0bfa1e29213/lib/ansible/plugins/action/debug.py#L55-L61
https://github.com/ansible/ansible/blob/4a57cba86d7de89aaae03e78ee5d4d87be79518e/lib/ansible/template/__init__.py#L570

That suggests using msg is better that using var as msg is not parsed and var is. I will change the code to use the var and will ping you to see if we can retry on your setup.

neelimamukiri:
I'm closing this as both msg and var are being evaluated and depending on the string contents and format, we may run into the ansible parsing issue. Adding debug code that can cause problems doesn't seem like a good idea. I've created https://github.com/contiv/ccn/issues/195 to track the debug scripts.

PR #307, opened 2017-01-19
roles/docker: support 1.12+ & delete docker.socket

### LINKED ISSUE #301: Update docker role to incorporate new changes for docker 1.12
With docker 1.12, docker.service removes the requires of requiring docker.socket. Update ansible scripts to handle this.

unclejack:
This PR removes the docker.socket and implements compatibility with Docker 1.12 (or newer). docker.socket has been removed upstream. Compatibility with Docker 1.11 is maintained.

Fixes #301


unclejack:
ping @gaurav-dalvi

gaurav-dalvi:
i dont have any comments !
@neelimamukiri may be able to review it !

unclejack:
@DivyaVavili @neelimamukiri Could you take a look when you have the time to do that, please?

roles/docker/templates/docker-svc.j2
```diff
@@ -1,17 +1,27 @@
...
+{% set docker_devmapper_opts = "--storage-opt dm.thinpooldev=/dev/mapper/contiv-dockerthin" %}
 {% else %}
-ExecStart=/usr/bin/docker daemon -H fd:// --cluster-store=etcd://localhost:{{ etcd_client_port1 }}
+{% set docker_devmapper_opts = "" %}
 {% endif %}
+
+{% if docker_version | version_compare('1.12', '>=') %}
+{% set docker_daemon_startup = "dockerd" %}
+{% else %}
+{% set docker_daemon_startup = "docker daemon" %}
```
neelimamukiri, line 18: Are we supporting the < 1.12 path? If yes, wont we need the docker-tcp tasks from above as well?

unclejack: @neelimamukiri: This was mostly a matter of how the systemd units were set up. They weren't absolutely necessary to run Docker. I've been running Docker for a long time without that setup.

neelimamukiri: Okay, that makes things simpler. Thanks!

neelimamukiri (approved):

roles/docker/templates/docker-svc.j2
dvavili: @unclejack: docker-tcp.socket is required to enable Docker remote API. docker client can still work without this, but we need to enable remote API for netplugin.

dvavili (changes requested):

roles/docker/templates/docker-svc.j2
unclejack: @DivyaVavili This is handled by lines 21 and 23 of this template. The UNIX socket and TCP socket options are passed directly to the Docker daemon as options on the command line. The docker-tcp.socket is no longer required.

It doesn't make sense to keep docker-tcp.socket for older versions. Older versions of Docker can also use this setup with `-H tcp://0.0.0.0:2375 -H unix:///var/run/docker.sock`. This was used to have socket activation and it was more trouble than it was worth. We can just use the latest recommended setup.

I've been using this setup for a long time. I haven't run into any kind of issue so far.

dvavili: OK. Sounds good. The changes LGTM in that case.

dvavili (approved):

unclejack:
I'll merge this now. We have two reviews.

Issue #304, opened 2017-01-13
start contiv related services on reboot


gaurav-dalvi:
Contiv related services I can think of
1: netplugin
2: netmaster
3: etcd
4: docker
5: swarm containers
(in the scope of swarm + contiv setup)

Current ansible scripts start them as systemd services. I tried adding 
```
- name: start etcd
  service: name=etcd enabled=yes
```
ref:
http://docs.ansible.com/ansible/service_module.html

https://github.com/contiv/ansible/blob/master/roles/etcd/tasks/main.yml#L43

But its not working.

PR #312, opened 2017-01-25
Set service state to enabled for contiv services

neelimamukiri:
Fixes https://github.com/contiv/ansible/issues/304
Also setting the Install section to allow enabling the services.

### DIFF
--- a/roles/contiv_network/tasks/main.yml
+++ b/roles/contiv_network/tasks/main.yml
@@ -54,7 +54,7 @@
   file: src=/usr/bin/contiv/netplugin/contrib/completion/bash/netctl dest=/etc/bash_completion.d/netctl state=link
 
 - name: start netplugin
-  shell: systemctl daemon-reload && systemctl start netplugin
+  service: name=netplugin daemon_reload=yes state=started enabled=yes
 
 - name: setup netmaster host alias
   lineinfile:
@@ -83,7 +83,7 @@
   when: run_as == "master"
 
 - name: start netmaster
-  shell: systemctl daemon-reload && systemctl start netmaster
+  service: name=netmaster daemon_reload=yes state=started enabled=yes
   when: run_as == "master"
 
 # FIXME: need to move the following tasks to correct role
--- a/roles/base/files/consul.service
+++ b/roles/base/files/consul.service
@@ -6,3 +6,6 @@ After=auditd.service systemd-user-sessions.service time-sync.target
 EnvironmentFile=/etc/default/consul
 ExecStart=/usr/bin/consul agent -server -data-dir /var/lib/consul $CONSUL_OPTS
 KillMode=process
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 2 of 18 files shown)


unclejack:
LGTM

@DivyaVavili @gaurav-dalvi: Could you review this when you have a chance, please?

roles/contiv_network/tasks/main.yml
```diff
@@ -82,8 +85,12 @@
   copy: src=netmaster.service dest=/etc/systemd/system/netmaster.service
   when: run_as == "master"
 
+- name: reload system config for netmaster
+  shell: systemctl daemon-reload
+  when: run_as == "master"
```
dvavili: Why is this change required?

neelimamukiri: Why do the reload before setting the state to started/enabled? If the service is not loaded, apparently setting it to enabled fails.

dvavili: OK. Could you use the daemon-reload option in service module rather than have a separate task for this?

(Ref: https://docs.ansible.com/ansible/systemd_module.html#options)

dvavili: Also, wouldn't we require the deamon-reload to happen even for etcd and aci-gw services?

dvavili (changes requested):

roles/contiv_network/tasks/main.yml
neelimamukiri: Sure, I can make that change as well. I wasn't really adding or changing the daemon-reload option, so I didn't check etcd and the other services. I will check and add that change along with this.

neelimamukiri: Given that etcd doesn't have the reload and still works fine, I wonder if it is required at all. I don't see any harm in adding it though, so will go ahead with it.

dvavili (approved):

dvavili:
Thanks for incorporating the review comments. Changes LGTM.

gaurav-dalvi:
LGTM. Very good and very quick @neelimamukiri . Keep it up.

gaurav-dalvi:
Thanks @neelimamukiri .
@unclejack suggested these cleanup steps . Is it possible to incorporate those into our ansible scripts ?

```
sudo rm -rf ansible .gen .ansible genInventoryFile.py
sudo docker kill -s 9 $(sudo docker ps -q)
sudo docker rm -f $(sudo docker ps -a -q)
sudo systemctl stop docker
sudo systemctl stop etcd
for i in $(mount | grep docker | awk '{ print $3 }'); do sudo umount $i || true; done
sudo umount /var/lib/docker/devicemapper || true
sudo yum remove -y docker-engine.x86_64 docker-engine-selinux.noarch
sudo rm -rf /usr/bin/etcd*
sudo rm -rf /var/lib/etcd*
sudo rm -rf /var/lib/docker
```

unclejack:
@gaurav-dalvi This PR only makes changes needed to allow the cluster to start after reboot. Cleanup will be a separate thing.

neelimamukiri:
@gaurav-dalvi, lets add a different issue to track the clean up - current task cleanup is only stopping services and not truly uninstalling anything.

PR #319, opened 2017-01-30
Adding auth_proxy role and install files

neelimamukiri:
The install_* and uninstall_* files allow triggering install and
cleanup actions for specific components. This allows the end user to
specify finer grained installation (only docker, only scheduler, only
contiv etc.)

### DIFF
--- a/roles/auth_proxy/README.md
+++ b/roles/auth_proxy/README.md
@@ -27,13 +27,3 @@ Example Playbook
   become: true
       roles:
         - { role: auth_proxy, auth_proxy_port: 10000, auth_proxy_datastore: etcd://netmaster:2379 }
-
-License
--------
-
-BSD
-
-Author Information
-------------------
-
-Please see contiv.io for more information.
--- a/roles/auth_proxy/README.md
+++ b/roles/auth_proxy/README.md
@@ -1,0 +1,39 @@
+Role Name
+=========
+
+Role to install Contiv API Proxy and UI
+
+Requirements
+------------
+
+Docker needs to be installed to run the auth proxy container.
+
+Role Variables
+--------------
+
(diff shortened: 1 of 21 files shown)


roles/auth_proxy/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+- hosts: netplugin-node
+  become: true
+      roles:
+        - { role: auth_proxy, auth_proxy_port: 10000, auth_proxy_datastore: etcd://netmaster:2379 }
+
+License
+-------
+
+BSD
```
jojimt: Why is this BSD?

uninstall_auth_proxy.yml
```diff
@@ -0,0 +1,23 @@
...
+    - include_vars: roles/{{ item }}/defaults/main.yml
+      with_items:
+        - "contiv_network"
+        - "contiv_storage"
+        - "swarm"
+        - "kubernetes"
+        - "ucp"
+        - "docker"
+        - "etcd"
+        - "auth_proxy"
```
jojimt: Should we be referring to contiv_storage and ucp here?

roles/auth_proxy/README.md
neelimamukiri: It started with the intent of making it an Ansible Galaxy role. I dont see any specific license for contiv_network. Should I remove the license field here?

uninstall_auth_proxy.yml
neelimamukiri: Ideally no. We have some cross references of defaults between docker, etcd and contiv. auth_proxy does not have any dependencies. I will remove the references to contiv_storage and ucp.

jojimt:
LGTM

PR #322, opened 2017-02-02
Make contiv network installation files local

neelimamukiri:
aci-gw and auth-proxy images, netplugin release, ovs rpm/deb files
will be loaded from local ansible files folders, if they exist.
If not we will continue to download from the urls or use docker
pull/docker run. This allows us to create a installation which is small
with external dependencies or larger but with no external dependencies
for contiv. Other roles are not modified as they are not part of the
core contiv installation.
Also changed the ansible module from service to systemd to allow the daemon_reload to take effect.


roles/contiv_network/tasks/aci_tasks.yml
```diff
@@ -1,7 +1,14 @@
 # These tasks are run when contiv_network_mode is set to aci
 
-- name: pull aci-gw container
-  shell: docker pull contiv/aci-gw
+# Load the aci-gw-image from local tar. Ignore any errors to handle the
+# case where the image is not built in
+- name: copy aci-gw image
+  copy: src={{ contiv_standalone_binaries }}/aci-gw-image.tar dest=/tmp/aci-gw-image.tar
+  ignore_errors: yes
```
jojimt: Can we make this conditional rather than using ignore_errors?

neelimamukiri: Is there a condition for source file exists? I was only able to find stat for destination file? Other option is to have it passed in an a parameter to the installation.

roles/auth_proxy/tasks/main.yml
```diff
@@ -1,6 +1,18 @@
 ---
 # tasks file for auth_proxy
 
+# Load the auth-proxy-image from local tar. Ignore any errors to handle the
+# case where the image is not built in
+- name: copy auth-proxy image
+  copy: src={{ auth_proxy_binaries }}/auth-proxy-image.tar dest=/tmp/auth-proxy-image.tar
+  when: auth_proxy_local_install == True
+  ignore_errors: yes
```
jojimt: Let's eliminate the ignore_errors for all such cases as we have the conditional var.

neelimamukiri: @jojimt, Done. The reason for ignore_errors was that all of these had fall back alternatives. If docker load or docker pull fail, the service can try to pull or complete the pull at a later time when we do a docker run. I do see your point that failing fast is good for cases where these really dont work, for example a bad image name etc.

jojimt:
LGTM

PR #324, opened 2017-02-03
Adding persistent storage backend for etcd

neelimamukiri:
This fixes https://github.com/contiv/ansible/issues/317

PR #325, opened 2017-02-05
Cleanup docker etcd state

### LINKED ISSUE #320: Docker role cleanup should remove any running containers
Same for swarm

neelimamukiri:
This change is made on top of https://github.com/contiv/ansible/pull/324, as we need the etcd state to cleanup. Fixes #320 and https://github.com/contiv/install/issues/12.

### DIFF
--- a/roles/docker/tasks/cleanup.yml
+++ b/roles/docker/tasks/cleanup.yml
@@ -1,6 +1,14 @@
 ---
 # This play contains tasks for cleaning up docker
 
+- name: remove all containers
+  shell: unset DOCKER_HOST && docker rm -f -v $(docker ps -a -q)
+  when: docker_reset_container_state == true
+
+- name: remove all images
+  shell: unset DOCKER_HOST && docker rmi -f $(docker images -q)
+  when: docker_reset_image_state == true
+
 - name: stop docker
   service: name=docker state=stopped
 
--- a/roles/docker/defaults/main.yml
+++ b/roles/docker/defaults/main.yml
@@ -14,3 +14,5 @@ docker_cs_engine_supported_versions_ubuntu:
     "Core": [""] # needed for redhat installations due to templating
     "Maipo": [""] # needed for redhat installations due to templating
 docker_mount_flags: "shared"
+docker_reset_container_state: false
+docker_reset_image_state: false
--- a/roles/etcd/defaults/main.yml
+++ b/roles/etcd/defaults/main.yml
@@ -11,3 +11,4 @@ etcd_version: "v2.3.7"
 etcd_heartbeat_interval: 1000
 etcd_election_timeout: 10000
 etcd_data_dir: /var/lib/etcd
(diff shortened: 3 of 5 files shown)


gaurav-dalvi:
LGTM

roles/docker/tasks/cleanup.yml
```diff
@@ -1,6 +1,14 @@
 ---
 # This play contains tasks for cleaning up docker
 
+- name: remove all containers
+  shell: unset DOCKER_HOST && docker rm -f -v $(docker ps -a -q)
```
gaurav-dalvi, line 5: Will this delete swamr-master and swarm-agent container ? 
So one I run uninstaller, it will cleanup etcd persistent storage and delete all docker containers and delete docker engine as well ?

gaurav-dalvi:
`sudo umount /var/lib/docker/devicemapper || true`
Do we need to do this as well ? @unclejack pointed that this needs to be done for every docker uninstallation.

roles/docker/tasks/cleanup.yml
neelimamukiri: Yes, it removes all containers, including swarm. As docker cleanup is run after swarm cleanup, this is fine. Uninstaller currently does not uninstall docker or any component. Cleanup only stops services. My guess is it was designed this way to allow multiple install/upgrades without having to deal with issues like the docker upgrade here.
We do need a set of uninstall tasks for all roles, to remove files, uninstall binaries etc. Thats a separate issue.

neelimamukiri:
This PR is only handling "state" cleanup - AFAIK thats only etcd & docker containers/images. Lets handle docker uninstall/upgrade in a separate issue.

gaurav-dalvi:
LGTM.

roles/docker/tasks/cleanup.yml
rhim, line 5: what's the rational for unset DOCKER_HOST? To target local docker engine?

roles/docker/tasks/cleanup.yml
```diff
@@ -1,6 +1,14 @@
...
+  when: docker_reset_container_state == true
+
+- name: remove all images
+  shell: unset DOCKER_HOST && docker rmi -f $(docker images -q)
```
rhim, line 9: Same question as above.

rhim (approved):
Minor nits.

roles/docker/tasks/cleanup.yml
neelimamukiri: Yes, the docker local service is being stopped here and we are trying to clean the containers locally as opposed to targeting any DOCKER_HOST specified in the shell defaults.

roles/docker/tasks/cleanup.yml
neelimamukiri: Same as above :)

PR #328, opened 2017-02-07
Fix #326, open the API proxy port

### LINKED ISSUE #326: open 10000 (auth_proxy port)
@JohnKDay just pointed that, our ansible scripts are not opening 10000 port .
We need to open it.

neelimamukiri:
Tested install/uninstall with the firewall enabled and checked the port is accessible with the iptables update.

### DIFF
--- a/roles/auth_proxy/defaults/main.yml
+++ b/roles/auth_proxy/defaults/main.yml
@@ -6,3 +6,4 @@ auth_proxy_key: "{{ contiv_certs }}/auth_proxy_key.pem"
 auth_proxy_datastore: "{{ cluster_store }}"
 auth_proxy_binaries: "/var/contiv_cache"
 auth_proxy_local_install: False
+auth_proxy_rule_comment: "Contiv auth proxy service"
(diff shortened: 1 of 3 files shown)


gaurav-dalvi:
Is there any specific reason to make auth_proxy as separate role ?

roles/auth_proxy/defaults/main.yml
```diff
@@ -4,3 +4,4 @@ contiv_certs: "/var/contiv/certs"
 auth_proxy_cert: "{{ contiv_certs }}/auth_proxy_cert.pem"
 auth_proxy_key: "{{ contiv_certs }}/auth_proxy_key.pem"
 auth_proxy_datastore: "{{ cluster_store }}"
+auth_proxy_rule_comment: "Contiv auth proxy service"
```
gaurav-dalvi: Can we add default for auth_proxy_port here ?

neelimamukiri: @gaurav-dalvi, auth_proxy_port was already there. https://github.com/neelimamukiri/ansible/blob/8c38141566ec3ab78a5a3ab91794dcc5c1b1c363/roles/auth_proxy/defaults/main.yml#L2

neelimamukiri:
The reason for having a separate auth_proxy role is to allow installations with only netplugin, without the auth proxy. Please see https://github.com/contiv/ccn/issues/145.

gaurav-dalvi:
LGTM

rhim (approved):
LGTM.

Issue #299, opened 2017-01-10
Make all the parameters of netplugin & netmaster configurable through the ansible scripts


neelimamukiri:
Currently there are many options to netplugin & netmaster.
See
https://github.com/contiv/netplugin/blob/master/netplugin/netd.go
https://github.com/contiv/netplugin/blob/master/netmaster/main.go

We need a way to specify them using the ansible role.
Some args are already being provided https://github.com/contiv/ansible/blob/master/roles/contiv_network/templates/netmaster.j2 and https://github.com/contiv/ansible/blob/master/roles/contiv_network/templates/netplugin.j2

We need to either add the others or have a generic parameter list string as an option.

To provide context, I need this as I am trying to run netmaster & etcd on a subset of the nodes and netplugin uses localhost: 2379 for the default cluster. Currently there is no arg for cluster_store in the ansible.

neelimamukiri:
Cluster store field was added in https://github.com/contiv/ansible/commit/1d7d2d725b3c23217ebda01face4af871d324021.
So netplugin does not need any additional parameter.
TODO: add listen-url (:9999) and dns-enable (false) parameters to the netmaster installation. I will add these.

rchirakk:
@neelimamukiri,  dns-enable (false) will be removed by https://github.com/contiv/netplugin/pull/661

neelimamukiri:
Thanks @rchirakk. So netmaster port is the only parameter left to be added.

rhim:
Since #331 isn't slated for beta, I am removing this issue from beta milestone.

PR #331, opened 2017-02-08
Passing netmaster_port paramter to netmaster initialization

neelimamukiri:
Fixes https://github.com/contiv/ansible/issues/299. We already had the netmaster_port being used as a parameter,  but it was only used to open the ports. Not adding it to netmaster as an option as well.


roles/contiv_network/templates/netmaster.j2
```diff
@@ -1 +1 @@
-NETMASTER_ARGS="--cluster-mode {{netplugin_mode}} -cluster-store {{cluster_store}} "
+NETMASTER_ARGS="--cluster-mode {{netplugin_mode}} -cluster-store {{cluster_store}} -listen-url :{{ netmaster_port }}"
```
jojimt, line 1: per @DivyaVavili we need some non-trivial changes before we can make the port configurable. We have https://github.com/contiv/netplugin/issues/705 to track this, but won't be done by beta time frame.

dvavili:
Closing this as it's already handled in #342

PR #332, opened 2017-02-11
Add ignore_errors option to cleanup

jojimt:


neelimamukiri:
@jojimt, we are ignoring the errors for all the included playbooks. Why would we have an error at the top level again, that is to be ignored? I'm wondering if this may affect cases where the playbook doesn't run at all due to hosts not being reachable, etc.

jojimt:
``we are ignoring the errors for all the included playbooks.``
We are not! We are currently ignoring only errors on the include task.

neelimamukiri (approved):
Thanks for the clarification!

PR #334, opened 2017-02-16
Add openstack repo

### LINKED ISSUE #335: Private box files got removed
Fixed in #334

neelimamukiri:
Fixes #335

### DIFF
--- a/roles/contiv_network/tasks/ovs.yml
+++ b/roles/contiv_network/tasks/ovs.yml
@@ -1,30 +1,40 @@
 ---
 # This role contains tasks for installing ovs
-
-- name: copy ovs file (redhat)
-  copy: src={{ contiv_standalone_binaries }}/openvswitch-2.3.1-2.el7.x86_64.rpm dest=/tmp/openvswitch-2.3.1-2.el7.x86_64.rpm 
-  when: ansible_os_family == "RedHat" and contiv_network_local_install == True
-
-- name: download ovs binaries (redhat)
+- name: get openstack kilo repo
   get_url:
-    validate_certs: "{{ validate_certs }}"
-    dest: "{{ item.dest }}"
-    url: "{{ item.url }}"
-  with_items:
-    - {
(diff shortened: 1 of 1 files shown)


neelimamukiri:
@unclejack has started a test run with contiv/netplugin, waiting on the results.

rhim:
Not a code bug per se, but a config issue, as privately held box files disappeared. This is the right way to fix for other binaries that we don't maintain.

unclejack:
LGTM

unclejack (approved):

roles/contiv_network/tasks/ovs.yml
```diff
@@ -1,30 +1,27 @@
...
   tags:
     - prebake-for-dev
 
-- name: install ovs (redhat)
-  yum: name=/tmp/openvswitch-2.3.1-2.el7.x86_64.rpm state=present
+- name: install openstack kilo rpm
+  yum: name=/tmp/rdo-release-kilo-2.noarch.rpm state=present
   when: ansible_os_family == "RedHat"
   tags:
     - prebake-for-dev
```
unclejack: We should disable the openstack-kilo repository by default. This has lead to the installation of urllib3 and another package from openstack-kilo.

We could use this to disable the repository by default immediately after the openstack kilo rpm is installed:
```
- name: disable openstack kilo repo
  ini_file:
    dest: /etc/yum.repos.d/rdo-release.repo
    section: openstack-kilo
    option: enabled
    value: 0
    mode: 0644
    backup: no
```

The ovs install would look like this:
```
- name: install ovs
  yum:
    pkg=openvswitch-2.5.0-2.el7
    state=present
    enablerepo=openstack-kilo
  tags:
    - prebake-for-dev
```

neelimamukiri: @unclejack, thanks for the info. So that explains why vagrantfile had the openstack kilo repo disabled by default. I'll make this change.

PR #336, opened 2017-02-21
roles/etcd: bump etcd to 2.3.8

unclejack:
This PR bumps etcd to 2.3.8. This version of etcd is compiled with Go 1.7.5. 2.3.8 includes a fix for data corruption when the disk is full.

https://github.com/coreos/etcd/releases/tag/v2.3.8


dseevr:
Good timing, because I'm updating Godeps in objdb and modelgen to use 2.3.8 😄

dseevr (approved):

unclejack:
I'll merge this now as I've tested the change. We can revert if we run into any issues.

PR #341, opened 2017-04-13
Update kilo repo to ocata

neelimamukiri:
Also adding the local ovs. Fixes https://github.com/contiv/install/issues/103

### DIFF
--- a/roles/contiv_network/tasks/ovs.yml
+++ b/roles/contiv_network/tasks/ovs.yml
@@ -1,29 +1,67 @@
 ---
 # This role contains tasks for installing ovs
-- name: get openstack kilo repo
+
+# Local install
+- name: copy ovs file (redhat)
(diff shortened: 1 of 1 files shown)


unclejack (approved):
LGTM

roles/contiv_network/tasks/ovs.yml
```diff
@@ -1,54 +1,92 @@
...
+  tags:
+    - prebake-for-dev
+
+# Remote install
+
+- name: get openstack ocata repo
   get_url:
-    url: https://repos.fedorapeople.org/repos/openstack/openstack-kilo/rdo-release-kilo-2.noarch.rpm
-    dest: /tmp/rdo-release-kilo-2.noarch.rpm
+    url: https://repos.fedorapeople.org/repos/openstack/openstack-ocata/rdo-release-ocata-2.noarch.rpm
```
gaurav-dalvi, line 43: what happens if after some time this repo is no longer valid or becomes EOL.

neelimamukiri: If we are using public repos we need to track changes to them. Given this is their latest version, it will hopefully be not too soon.

PR #344, opened 2017-05-16
Added support for v2plugin install

gkvijay:
- first set of changes for ansible task for v2plugin install

### DIFF
--- a/roles/v2plugin/defaults/main.yml
+++ b/roles/v2plugin/defaults/main.yml
@@ -1,0 +1,37 @@
+---
+
+# Role defaults for v2plugin
+# NOTE: Role defaults have a lower priority than inventory vars.
+# Include variables which need to be overridden by inventory vars here.
+
+contiv_v2plugin_image: "contiv/v2plugin:0.1"
+
+contiv_network_mode: "standalone" # Accepted values: standalone, aci
+netplugin_mode: "docker" # Accepted values: docker, kubernetes
+fwd_mode: "bridge" #Accepted values: bridge , routing
+
+cluster_store: "etcd://127.0.0.1:2379"
+contiv_standalone_binaries: "/var/contiv_cache"
+ofnet_master_port: 9001
+ofnet_agent_port1: 9002
+ofnet_agent_port2: 9003
+netmaster_port: 9999
+gobgp_grpc_port: 8080
+bgp_port: 179
+# Since there is a conflict on vxlan port used by Docker overlay (ingress)
+# contiv needs to use a different port than the standard 4789
+vxlan_port: 8472
+netplugin_rule_comment: "contiv network traffic"
+
+aci_gw_image: "contiv/aci-gw"
+
+# contiv tar files to get netctl binary
+contiv_network_version: "v0.1-11-30-2016.20-08-20.UTC"
+contiv_network_tar_file: "netplugin-{{ contiv_network_version }}.tar.bz2"
+contiv_network_src_tar_file: "{{ contiv_standalone_binaries }}/{{ contiv_network_tar_file }}"
+contiv_network_src_file: "https://github.com/contiv/netplugin/releases/download/{{ contiv_network_version }}/{{ contiv_network_tar_file }}"
+contiv_network_dest_file: "/tmp/{{ contiv_network_tar_file }}"
+
+apic_epg_bridge_domain: "not_specified"
+apic_contracts_unrestricted_mode: "no"
+
--- a/roles/v2plugin/tasks/main.yml
+++ b/roles/v2plugin/tasks/main.yml
@@ -1,0 +1,67 @@
+---
+# This role contains tasks for starting docker plugin service
+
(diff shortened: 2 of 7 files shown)


roles/v2plugin/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+- name: setup netmaster host alias
+  lineinfile:
+    dest: /etc/hosts
+    line: "{{ service_vip }} netmaster"
+    regexp: " netmaster$"
+    state: present
+
+- name: install v2plugin on {{ run_as }} nodes
+  shell: >
+    /usr/bin/docker plugin install --grant-all-permissions {{ contiv_v2plugin_image }} ctrl_ip={{ node_addr }} control_url={{ node_addr }}:9999 vxlan_port=8472 iflist={{ netplugin_if }} plugin_name={{contiv_v2plugin_image}} cluster_store={{cluster_store}} plugin_role={{ run_as }}
```
neelimamukiri: Can you move these ports to defaults as variables. 9999 already has netmaster_port var.

roles/v2plugin/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+    validate_certs: "{{ validate_certs }}"
+    url: "{{ contiv_network_src_file }}"
+    dest: "{{ contiv_network_dest_file }}" 
+  when: (contiv_network_local_install == False)
+
+- name: ensure netplugin directory exists
+  file: path=/usr/bin/contiv/netplugin state=directory
+
+- name: extract binaries from tar file
+  shell: tar vxjf {{ contiv_network_dest_file }}
```
neelimamukiri: Only netctl is required here, right? 
Use something like `tar oxf {{ contiv_network_dest_file }} netctl`

roles/v2plugin/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+
+# setting global info needs netmaster to be ready
+- name: wait for netmaster to be ready
+  action: shell netctl tenant ls
+  register: tenant_result
+  until: tenant_result.stdout.find("default") != -1
+  retries: 9
+  delay: 10
+
+- include: aci_tasks.yml
```
neelimamukiri, line 61: Can this point to the one in contiv_network to avoid duplication. Not sure how it works in the current form as the file should be missing.

gkvijay: Copied aci_tasks.yml in the v2plugin role as well

gkvijay: @neelimamukiri PTAL

neelimamukiri (approved):
LGTM. Please raise an issue to track refactoring the aci_gw stuff.

roles/v2plugin/tasks/main.yml
```diff
@@ -0,0 +1,66 @@
...
+  shell: >
+      ( iptables -L INPUT | grep "{{ netplugin_rule_comment }} ({{ item }})" ) || \
+      iptables -I INPUT 1 -p tcp --dport {{ item }} -j ACCEPT -m comment --comment "{{ netplugin_rule_comment }} ({{ item }})"
+  with_items:
+    - "{{ ofnet_master_port }}"
+    - "{{ ofnet_agent_port1 }}"
+    - "{{ ofnet_agent_port2 }}"
+    - "{{ netmaster_port }}"
+    - "{{ gobgp_grpc_port }}"
+    - "{{ bgp_port }}"
```
dvavili, line 14: We have to open up the vxlan port as well. Or is it opened up as part of some other role?

roles/v2plugin/defaults/main.yml
```diff
@@ -0,0 +1,37 @@
...
+contiv_standalone_binaries: "/var/contiv_cache"
+ofnet_master_port: 9001
+ofnet_agent_port1: 9002
+ofnet_agent_port2: 9003
+netmaster_port: 9999
+gobgp_grpc_port: 8080
+bgp_port: 179
+# Since there is a conflict on vxlan port used by Docker overlay (ingress)
+# contiv needs to use a different port than the standard 4789
+vxlan_port: 8472
```
dvavili, line 23: Specific reason to choose this port number? Why not stick with the same port as what contiv_network chooses? Just that it is easier to debug in different modes if the same port is chosen.

dvavili (commented):
As part of the refactoring, also consider moving the v2plugin tasks into a separate task file that can be included, rather than having a separate role. In that case, we can minimize the effort to make sure updates in contiv_network and v2plugin are synchronized (as the case may be).

roles/v2plugin/defaults/main.yml
gkvijay: cannot use the same port as there is a conflict

roles/v2plugin/tasks/main.yml
gkvijay: added

PR #347, opened 2017-05-19
k8s_tasks.yml: make some changes needed for k8s

unclejack:
This PR brings in more changes needed for the Kubernetes installation.

### DIFF
--- a/roles/contiv_network/tasks/k8s_tasks.yml
+++ b/roles/contiv_network/tasks/k8s_tasks.yml
@@ -12,13 +12,13 @@
     mode: u=rwx,g=rx,o=rx
     remote_src: yes
     force: yes
-  notify: restart kubernetes
+  #notify: restart kubernetes
 
 - name: copy contiv k8s plugin config file
   copy:
     src: contiv_cni.conf
     dest: "{{ k8s_net_plugin_config_dir }}/contiv_cni.conf"
-  notify: restart kubernetes
+  #notify: restart kubernetes
 
 - name: ensure contiv plugin config dir exists
   file:


roles/contiv_network/tasks/k8s_tasks.yml
```diff
@@ -5,20 +5,30 @@
...
+
 - name: copy contiv cni bin to k8s plugin dir
   copy:
     src: "/usr/bin/contiv/netplugin/contivk8s"
     dest: "{{ k8s_net_plugin_bin_dir }}/contivk8s"
     mode: u=rwx,g=rx,o=rx
     remote_src: yes
     force: yes
-  notify: restart kubernetes
+  #notify: restart kubernetes
```
rhim, line 15: Is this for commenting out?

unclejack: @rhim: We'll have to disable this for now. We won't support the installation of kubernetes with these playbooks for the time being.

gaurav-dalvi:
Could you please explain why do we need this change ? 
title of the PR does not give any idea.

unclejack:
This is work in progress. I'm investigating some issues.

roles/contiv_network/tasks/k8s_tasks.yml
neelimamukiri: @unclejack , can you raise a CCN issue to track re-enabling this for CI/CD.

roles/contiv_network/tasks/k8s_tasks.yml
```diff
@@ -5,20 +5,30 @@
...
 
+- name: ensure kubernetes dirs exist
+  file:
+    path: "{{ item }}"
+    recurse: yes
+    state: directory
+  with_items:
+    - "{{ k8s_manifests_dir }}"
+    - "{{ k8s_cert_dir }}"
+    - "{{ k8s_net_plugin_config_dir }}"
```
neelimamukiri: Netplugin dir part is required. Why do we need the k8s_manifests & certs here? Netplugin wont work in an install which doesnt have these folders. And by having these created, we increase the chance that some one misconfigures the folder paths. As we depend on the content of these folders, and we dont create them, its better to fail early if they are not there.

unclejack: I've removed this change.

neelimamukiri (approved):

PR #355, opened 2017-06-08
Make netctl URL configurable

neelimamukiri:


dvavili (commented):
This is not a very clean way to do it. Say, the listen_url is configured to :9970, netctl_url also has to be configured to http://netmaster:9970. And if listen_url is configured with an explicit IP as in <IP>:<Port> netctl_url should be pointing to the same <IP>:<Port>. Is this ensured to be in sync?

neelimamukiri:
Yes, both these values have to be specified in sync, but having netctl_url separate allows the use to select either listen or control URL as required.

PR #360, opened 2017-06-21
Allow configuring netmaster URL

neelimamukiri:
Also cleaning up the v2plugin role

Removing /etc/hosts entries for netmaster as we want to allow netmaster to be localhost for master nodes and to be specified as an external load balanced auth_proxy VIP for accessing netmaster from anywhere else.


rhim (approved):
LGTM from change side. CI needs to re-run.

gkvijay (approved):

PR #361, opened 2017-07-19
Commenting out code for creating 2nd interface on VMs. Not needed. Ch…

dojacobs:
…anging machine to t2.micro to save money.

### DIFF
--- a/aws.tf
+++ b/aws.tf
@@ -92,7 +92,7 @@ variable "aws_ami" {
 }
 
 variable "aws_instance_type" {
-    default = "t2.large"
+    default = "t2.micro"
 }
 
 # how many VMs to spin up. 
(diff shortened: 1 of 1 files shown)


neelimamukiri (approved):
LGTM. Can we may be split this into two - aws.tf and aws-large.tf or something similar, instead of having commented out and stale code?

dojacobs:
Since this is only used by Jenkins, I don't see a point in having a "small" and "large" version.  If you would like, I could make machine type  an external variable that you specify in the aws.tfvars file or it could be specified on the command line when you do the terraform apply.  Otherwise it's just one line inside the script to change the machine type that is used.

neelimamukiri:
@dojacobs, in that case why not just delete the commented lines?

dojacobs:
@rhim's thinking was that we might need them in the future.  It's no big deal either way.  What is the normal process?

dojacobs:
Updated file with the commented lines deleted.

dseevr (approved):

dseevr:
@contiv/maintainers-deployment PTAL

PR #365, opened 2017-09-14
Update IPTable rules and etcd template  to support persistent across reboot .    

g1rana:


roles/etcd/templates/etcd.j2
```diff
@@ -131,10 +131,11 @@ init_update_cluster() {
     # on master nodes, if the cluster is being initialized for first time then initialize it
     peers=""
     get_peers "{{ get_peer_addr() }}"
+    rm -rf ${ETCD_DATA_DIR}/wal/*
```
srampal: Will this remove the etcd data directory even for the case of cluster getting updated ? That may not be what we want.

roles/etcd/templates/etcd.j2
```diff
@@ -131,10 +131,11 @@ init_update_cluster() {
...
     if [ "${peers}" == "" ]; then
         echo "==> no peers configured or rest of the etcd cluster is unhealthy, trying to init the cluster"
         export ETCD_INITIAL_CLUSTER_STATE=new
-        export ETCD_INITIAL_CLUSTER="
+	 export ETCD_INITIAL_CLUSTER="
```
srampal: Minor typo, added an unnecessary extra space indentation.

g1rana: ok

rchirakk (approved):
LGTM

roles/etcd/templates/etcd.j2
g1rana: No, it won't delete data directory on update. it's only cleanup directory whenever ETCD node/container will  start or restart. During debugging it has been found that restart node/container was not able to join cluster.
This work around would let the restart ETCD node to rejoin the cluster again. Looks like this is know issue in ETCD . Found following - 
“etcd process hanging after removed from membership and restart #7512”

srampal: ok, sounds good. 
LGTM

g1rana:
@rchirakk @srampal 
Please review it again and merge it  if possible

rhim:
Fixes https://github.com/contiv/netplugin/issues/963

PR #370, opened 2017-09-28
download the OVS 2.5 RPM & install it directly

unclejack:
This PR removes the dependency on the ocata openstack repository. This repository was used to install openvswitch 2.5.0. This OVS package was removed and newer versions are now available. We can't rely on this newer version of the package because it could get upgraded unexpectedly and it hasn't been tested with netplugin. Newer packages of OVS 2.6 have been provided via this repository.

This PR modifies the code to retrieve the same package we were using before from a CentOS server. This fixes the deployment failures we were running into which were caused by the removal of the OVS 2.5.0 RPM from the ocata repository.

This was tested with contiv/install.

### DIFF
--- a/roles/contiv_network/tasks/ovs.yml
+++ b/roles/contiv_network/tasks/ovs.yml
@@ -38,38 +38,28 @@
 
 # Remote install
 
-- name: get openstack ocata repo
+- name: get the openvswitch RPM from centos.org
   get_url:
-    url: https://repos.fedorapeople.org/repos/openstack/openstack-ocata/rdo-release-ocata-2.noarch.rpm
-    dest: /tmp/rdo-release-ocata-2.noarch.rpm
-    validate_certs: false
+    url: https://cbs.centos.org/kojifiles/packages/openvswitch/2.5.0/2.el7/x86_64/openvswitch-2.5.0-2.el7.x86_64.rpm
+    dest: /tmp/openvswitch-2.5.0-2.el7.x86_64.rpm
+    checksum: sha256:2ff007391ede80f9887495526e8e1c45d1ce746b2b97efe8e258279ca3a75467
   when: ansible_os_family == "RedHat" and contiv_network_local_install == False
   tags:
     - prebake-for-dev
+  register: ovs_rpm_download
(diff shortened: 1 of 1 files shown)


roles/contiv_network/tasks/ovs.yml
```diff
@@ -38,38 +38,16 @@
 
 # Remote install
 
-- name: get openstack ocata repo
+- name: get the openvswitch RPM
   get_url:
-    url: https://repos.fedorapeople.org/repos/openstack/openstack-ocata/rdo-release-ocata-2.noarch.rpm
-    dest: /tmp/rdo-release-ocata-2.noarch.rpm
-    validate_certs: false
+    url: https://cbs.centos.org/kojifiles/packages/openvswitch/2.5.0/2.el7/x86_64/openvswitch-2.5.0-2.el7.x86_64.rpm
```
rhim, line 43: There is no guarantee that this repo/personal folder won't disappear tomorrow.

We need to host the file in our own repo. We might already have this as part of openstack repo that Metacloud folks host. Let's use that repo to fix this issue the right way.

rhim (changes requested):

roles/contiv_network/tasks/ovs.yml
kahou82: Currently we don't have a production repo for public access; All the metacloud production repo requires VPN. In the meantime, @rhim suggests to put OVS 2.5 rpm in github to mitigate the issue. 

For the longer term, we need to setup our public accessible repo to lock down the dependency versions. I have ticket created to track this workstream.

unclejack: I've updated the PR to have two locations. It'll try to fetch the RPM from this location first and fall back to GitHub. This should be fine. I've also included a file hash to make sure it's the same file.

unclejack:
I've added another download location. The second download location is a fallback location.

This should address the issue for the time being. I've tested the main download location and the fallback.

PTAL @kahou82 @rhim

kahou82:
👍

dseevr:
👍

PR #373, opened 2017-10-11
DO NOT MERGE: Fix iptables persistence after reboot

kahou82:
Currently, etcd iptables rule doesn't persist after reboot. This is
because we have to enable the iptables service. Otherwise, iptables
service will not load after reboot

### DIFF
--- a/roles/etcd/tasks/main.yml
+++ b/roles/etcd/tasks/main.yml
@@ -1,3 +1,4 @@
+[metacloud@mcp1 ansible]$ cat roles/etcd/tasks/main.yml
 ---
 # This role contains tasks for configuring and starting etcd service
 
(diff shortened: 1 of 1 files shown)


roles/etcd/tasks/main.yml
```diff
@@ -26,14 +27,20 @@
   yum: pkg={{ item }} state=present
   with_items:
    - iptables-services
-  when: ansible_distribution == 'CentOS' or ansible_distribution == 'Red Hat Enterprise Linux'
+  when: ansible_distribution == 'CentOS' or ansible_distribution == 'RedHat'
```
tiewei, line 29: should be able to use `ansible_os_family` instead

roles/etcd/tasks/main.yml
```diff
@@ -1,3 +1,4 @@
+[metacloud@mcp1 ansible]$ cat roles/etcd/tasks/main.yml
```
tiewei: what's this ?

roles/etcd/tasks/main.yml
```diff
@@ -26,14 +27,20 @@
...
 
-- name: Install iptables-persistent  
+- name: Install iptables-persistent
   apt: name={{item}} state=installed
   with_items:
        - iptables-persistent
   when: ansible_distribution == 'Debian' or ansible_distribution == 'Ubuntu'
 
+- name: Enable iptables service
+  command: systemctl enable iptables
```
tiewei, line 38: use service module instead

roles/etcd/tasks/main.yml
```diff
@@ -26,14 +27,20 @@
...
+
+- name: Start iptables service
+  command: systemctl start iptables
```
tiewei, line 41: ditto

roles/etcd/tasks/main.yml
```diff
@@ -45,10 +51,10 @@
     - "{{ etcd_peer_port1 }}"
     - "{{ etcd_peer_port2 }}"
 
-#  Save rules into /etc/sysconfig/iptables file for restoring rules on boot 
+#  Save rules into /etc/sysconfig/iptables file for restoring rules on boot
 - name: Save iptables
   command: service iptables save
```
tiewei, line 56: I think we should use systemctl instead ... , but this may still work on older version of rhel

PR #374, opened 2017-10-12
In swarm-mode cluster, contiv iptables rules getting lost after reboot

g1rana:
This issue is seen in CVD testing  where in swarm-mode cluster, on rebooting the node, contiv container are not coming up since contiv iptables rules are not getting restore. CVD setup use REHL as base OS and firewalld is default application. With firewalld , iptables-service can not be enable therefore  we have add self restoring service to restore iptables rules  for contiv container. With this changes, initially ansible will install new service "contivRule.service" which  save all contiv rules into "/etc/contiv/rules.conf" file  after intalling all contiv container. During reboot, this new service will replay all iptables rules which has added before reboot from "/etc/contiv/rules.conf"
Following changes has been made for this PR : 
1. add contiv tag into all contiv specific iptables rules.
2. add new service "contivRule.service" to restore/replay(after reboot) all contiv rules into files. 
3. update v2plugin.sh file to aviod exist on "docker plugin enable " failure 
@dseevr @rchirakk @neelimamukiri @kahou82  @srampal please review it


roles/auth_proxy/files/contivRule.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=run contiv rule service 
+
```
dseevr, line 3: Do we want a `After=network.target` here?  I'm not sure at what point in systemd's startup we can guarantee the netfilter drivers are loaded

roles/contiv_network/files/v2plugin.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -euxo pipefail
+set -e
```
dseevr: Why this change?  We generally want to keep these things enabled to make our scripts harder to mess up 😄

roles/auth_proxy/files/contivRule.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e
+IFS="
+"
+rules_file="/etc/contiv/rules.conf"
+if [ -f "$rules_file" ]
+then
+    while read line; do
+        eval iptables $line
```
tiewei, line 9: this is dangerous, if there is anything or anyone changed the file by accident, you may lose root access to the box.
instead i would suggest to put logic in this file about 
1. if rules exists, skip
2. if rule missed, add the rule
And rules can be in different file that templated.
The benefit of doing this are, you don't have to use shell to manipulate iptables in ansible anymore, all you need is to enable and start this service, and you don't have to manage the iptables in serf start up script either, because you can manage the service dependency between these two services.

roles/auth_proxy/files/contivRule.service
tiewei: and also before docker or whatever service that requires the rules in place

kahou82:
I want to clarify that the right fix is to have another role which takes care of the iptables rules and those rules must be in a conf file. Then that role will take care of the idempotent. This way will be much cleaner.

The current fix can temporary fix the CVD issue but it is not clean enough as there is no idempotent in the role so far.

roles/consul/tasks/main.yml
```diff
@@ -39,8 +39,8 @@
 
 - name: setup TCP iptables for consul
   shell: >
-      ( iptables -L INPUT | grep -q 'consul tcp ({{ item }})' ) || \
-      iptables -I INPUT 1 -p tcp --dport {{ item }} -j ACCEPT -m comment --comment "consul tcp ({{ item }})" 
+      ( iptables -L INPUT | grep -q 'contiv_consul tcp ({{ item }})' ) || \
+      iptables -I INPUT 1 -p tcp --dport {{ item }} -j ACCEPT -m comment --comment "contiv_consul tcp ({{ item }})"
```
tiewei, line 43: there is `iptables --check` in rhel 7.4 that helps you check if a rule exists

roles/auth_proxy/files/contivRule.service
kahou82: If you have this service...just wonder why we have all the rules in the roles? We can do all the checking here, right? Like this:

1. During installation, generate all the rules in a folder
2. Install this service before anything relate to container (Docker, etc)
3. When this service start, the logic inside will check if we need to add the rules.

tiewei: indeed

g1rana: @dseevr I believe, netfilter is part of kernel and should always up with kernel itself before any services anyway  it would be safe to include "After=network.target" .
@tiewei , we don't need to taken care for docker rules since it would be taken care by docker-swarm/docker engine 
@kahou82 Initially i was thinking do same but couldn't quick logic to place it . therefore use existing role infra to put my changes.  Would need some help if want to go   above suggested solution

roles/consul/tasks/main.yml
kahou82: Does --check avaliable to all platform?

tiewei: not all only exists on newer distribution

roles/auth_proxy/files/contivRule.sh
g1rana: @tiewei , i  understand your concern and would need your help to implement your suggested solution. I am still in learning phase of scripting part :) and very familiar their proper uses in production environment.

roles/consul/tasks/main.yml
g1rana: I  just changed the tag name in existing rule code. Don't want  to change current logic for iptables. Also this current logic  would work for all OS distribution. 
We can open new PR for your suggested approach and fix after CVD . This approach require to change in all available rules.

g1rana:
Here is Unit Test ; 
Case 1 : 
1. Bring up docker-swarm cluster by using docker-swarm installer
2. Bring up contiv service by installing contiv installer 
3 . Verify everything application up and running 
4. reload cluster node using vagrant reload node 
5. confirm , docker container , etcd cluster and netplugin up and running after reboot 
===
Here is result : 
jerana@contiv271:~/git/src/github.com/contiv/install/cluster$ vagrant reload swarm-mode-master
==> swarm-mode-master: Attempting graceful shutdown of VM...
==> swarm-mode-master: Clearing any previously set forwarded ports...
==> swarm-mode-master: Fixed port collision for 22 => 2222. Now on port 2202.
==> swarm-mode-master: Clearing any previously set network interfaces...
==> swarm-mode-master: Preparing network interfaces based on configuration...
    swarm-mode-master: Adapter 1: nat
[Unit]
Description=run contiv rule service
    swarm-mode-master: Adapter 2: hostonly
    swarm-mode-master: Adapter 3: intnet
[Unit]
==> swarm-mode-master: Forwarding ports...
    swarm-mode-master: 22 (guest) => 2202 (host) (adapter 1)
==> swarm-mode-master: Running 'pre-boot' VM customizations...
==> swarm-mode-master: Booting VM...
==> swarm-mode-master: Waiting for machine to boot. This may take a few minutes...
    swarm-mode-master: SSH address: 127.0.0.1:2202
    swarm-mode-master: SSH username: vagrant
    swarm-mode-master: SSH auth method: private key
    swarm-mode-master: Warning: Remote connection disconnect. Retrying...
    swarm-mode-master: Warning: Remote connection disconnect. Retrying...
==> swarm-mode-master: Machine booted and ready!
==> swarm-mode-master: Checking for guest additions in VM...
    swarm-mode-master: The guest additions on this VM do not match the installed version of
    swarm-mode-master: VirtualBox! In most cases this is fine, but in rare cases it can
    swarm-mode-master: prevent things such as shared folders from working properly. If you see
    swarm-mode-master: shared folder errors, please make sure the guest additions within the
    swarm-mode-master: virtual machine match the version of VirtualBox you have installed on
    swarm-mode-master: your host and reload your VM.
    swarm-mode-master:
    swarm-mode-master: Guest Additions Version: 5.1.16
    swarm-mode-master: VirtualBox Version: 5.0
==> swarm-mode-master: Setting hostname...
==> swarm-mode-master: Configuring and enabling network interfaces...
==> swarm-mode-master: Mounting shared folders...
    swarm-mode-master: /shared => /home/jerana/git/src/github.com/contiv/install/cluster/export
    swarm-mode-master: /vagrant => /home/jerana/git/src/github.com/contiv/install/cluster
==> swarm-mode-master: Machine already provisioned. Run `vagrant provision` or use the `--provision`
==> swarm-mode-master: flag to force provisioning. Provisioners marked to run always will still run.
jerana@contiv271:~/git/src/github.com/contiv/install/cluster$ vagrant ssh swarm-mode-master
Last login: Fri Oct 13 19:24:07 2017 from 10.0.2.2
[vagrant@swarm-mode-master ~]$ docker ps
CONTAINER ID        IMAGE                        COMMAND                  CREATED             STATUS              PORTS               NAMES
05df5b214278        contiv/auth_proxy:1.0.3      "./auth_proxy --tl..."   39 seconds ago      Up 38 seconds                           auth-proxy
e8a7a7c758af        quay.io/coreos/etcd:v2.3.8   "/etcd"                  51 seconds ago      Up 49 seconds                           contiv_etcd
[vagrant@swarm-mode-master ~]$ sudo docker-runc list
ID                                                                 PID         STATUS      BUNDLE                                                                                       CREATED                          OWNER
05df5b214278f1e40c4a5605bada9f4aaa50cd7f6453a911fea63784cc1636b5   4842        running     /run/docker/libcontainerd/05df5b214278f1e40c4a5605bada9f4aaa50cd7f6453a911fea63784cc1636b5   2017-10-13T19:52:29.898475561Z   root
7ba8318cf26f50214cab7a306bcaedfa9561bd3c8bd4078ce5d57f88a002e192   4979        running     /run/docker/libcontainerd/7ba8318cf26f50214cab7a306bcaedfa9561bd3c8bd4078ce5d57f88a002e192   2017-10-13T19:52:40.930756609Z   root
e8a7a7c758afc4ef84ea517e4ac5a9062efb08f8f00972a28240f251b2fc6121   4573        running     /run/docker/libcontainerd/e8a7a7c758afc4ef84ea517e4ac5a9062efb08f8f00972a28240f251b2fc6121   2017-10-13T19:52:18.60359369Z    root
==

kahou82:
We need to revisit the iptables rules generation and handling in the future. Other than that, LGTM. 👍

tiewei (approved):
agree as a dirty fix to make it from not working to working, we will revisit it

roles/auth_proxy/files/contivRule.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e
```
chrisplo, line 2: for future, `set -euo pipefail`

dseevr (approved):

roles/contiv_network/files/v2plugin.sh
```diff
@@ -5,8 +5,7 @@ do
         ID="$(docker plugin ls  | awk '/contiv/ {print $2}')"
         STATUS="$(docker plugin ls  | awk '{print $8}')"
         if [ $STATUS != true ]; then
-                docker plugin enable $ID
-                sleep 1
+                docker plugin enable $ID || sleep 1
```
chrisplo, line 8: this is in https://github.com/contiv/ansible/pull/375 (but that merged already, all good)

roles/auth_proxy/files/contivRule.sh
```diff
@@ -0,0 +1,15 @@
...
+IFS="
```
chrisplo, line 3: don't get this, `read` will read the entire line already

roles/auth_proxy/files/contivRule.sh
chrisplo: the eval seems unnecessary, I don't see any reason to eval vs just running `iptables $line`

chrisplo:
not blocking, just added to list of TODOs

roles/auth_proxy/files/contivRule.sh
dseevr: The `eval` is needed because there are spaces in the $var and they will get quoted

Quick example:

```sh
export ARGS="-l -h"
ls $ARGS # fails
eval ls $ARGS # succeeds
```

PR #376, opened 2017-10-24
Rewrite etcd role for proper lifecycle

chrisplo:
Previously, docker could not be restarted on a node in a cluster and
have v2plugin remaining working due to a conflict because the container
was never stopped, the container name killed now matches that start name
for the container.

Previously, docker could not be shut down on multiple nodes without
getting etcd in a state where it could not start because the start/stop
script was dynamically managing the cluster membership as part of
service start/stop state.

Etcd is a HA database, starting or stopping should not depend on a
single node, the old code checked the first node for peers and used
that to determine if the cluster was being initialized or was already
active.  Instead, just configure etcd statically with all the member
nodes in configuration.

To improve readability, much of the old etcd start/stop scripts
template to produce urls for etcd configuration was jinja macros,
redid that as role default variables using composition instead.

Introduced some naming for the etcd cluster that drives systemd service
naming

etcd member names are by default based on the hostname instead of
numbers that can be out of order with the machines.

Added molecule testing for role, including gitignore, pip requirements
for dependencies, and a virtualbox configuration for testing etcd
install.

Drive-by:
* Add names to some plays and tasks to improve output
* Removed etcd peer port 7001 as it is not used

Signed-off-by: Chris Plock <chrisplo@cisco.com>


unclejack:
This is on hold as far as I'm concerned. The testing needs to be discussed and reviewed.

roles/etcd/molecule/default/molecule.yml
```diff
@@ -0,0 +1,57 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: vagrant
+  provider:
+    name: virtualbox
+platforms:
+    - name: etcd1
+      box: contiv/centos73
```
unclejack: This needs to be pinned to a specific version.

unclejack:
The Makefile needs a new target to run these tests (e.g. test-molecule).

roles/etcd/molecule/default/molecule.yml
chrisplo: done

chrisplo:
molecule converge is good:
```
    PLAY RECAP *********************************************************************
    etcd-proxy                 : ok=11   changed=9    unreachable=0    failed=0
    etcd1                      : ok=11   changed=9    unreachable=0    failed=0
    etcd2                      : ok=11   changed=9    unreachable=0    failed=0
    etcd3                      : ok=11   changed=9    unreachable=0    failed=0
```

chrisplo:
and install seems good: `Install SUCCESS`
`cluster is healthy`
`eb6723dcb5fd        contiv/v2plugin:1.0.3   Contiv network plugin for Docker   true`

```
[vagrant@swarm-mode-worker0 ~]$ sudo systemctl restart docker
[vagrant@swarm-mode-worker0 ~]$ docker plugin ls
ID                  NAME                    DESCRIPTION                        ENABLED
653be47da576        contiv/v2plugin:1.0.3   Contiv network plugin for Docker   false

.... time passes .....

[vagrant@swarm-mode-worker0 ~]$ docker plugin ls
ID                  NAME                    DESCRIPTION                        ENABLED
653be47da576        contiv/v2plugin:1.0.3   Contiv network plugin for Docker   true
```

```
[vagrant@swarm-mode-master ~]$ sudo systemctl restart docker
[vagrant@swarm-mode-master ~]$ docker plugin ls
ID                  NAME                    DESCRIPTION                        ENABLED
eb6723dcb5fd        contiv/v2plugin:1.0.3   Contiv network plugin for Docker   false
[vagrant@swarm-mode-master ~]$ systemctl status etcd
● etcd-contiv.service - Etcd for contiv
   Loaded: loaded (/etc/systemd/system/etcd-contiv.service; enabled; vendor preset: disabled)
   Active: active (running) since Tue 2017-10-24 21:39:00 UTC; 38s ago
  Process: 23288 ExecStop=/usr/bin/docker kill etcd-contiv (code=exited, status=1/FAILURE)
 Main PID: 23718 (docker)
   Memory: 4.6M
   CGroup: /system.slice/etcd-contiv.service
           └─23718 /usr/bin/docker run -t --rm --net=host --env-file /etc/etcd_env -v /var/lib/etcd/data:/var/lib/etcd/data -v /var/lib/etcd/wal:/var/lib/etcd/wal --name etcd-contiv quay.io/coreos/etcd:v2.3.8
[vagrant@swarm-mode-master ~]$ docker plugin ls
ID                  NAME                    DESCRIPTION                        ENABLED
eb6723dcb5fd        contiv/v2plugin:1.0.3   Contiv network plugin for Docker   true
```

chrisplo:
After shutting down docker on all nodes, docker can be started on all three, etcd started on all three, and cluster is healthy.

Had to start etcd manually due to a long docker down time, and with etcd as restart=always you quickly get to a state of `failed (Result: start-limit)` which requires manual starting.

We need to avoid the start-limit, for example no restart is required if docker is not running

g1rana:
just curious, Have you run reload node (both master  and worker)  test too.

chrisplo:
yes, did a three node master + 1 worker and long as you restart and not stop . . wait . . start etcd then v2plugin will restore, if you wait, you have to start etcd manually (will work on that)

roles/etcd/tasks/main.yml
```diff
@@ -1,61 +1,74 @@
...
+      shell: docker pull quay.io/coreos/etcd:{{ etcd_version }}
+      tags:
+        - prebake-for-dev
+
+    # Make Iptables rule persistent across reboot
+    - name: Install iptable-service package
+      yum: pkg={{ item }} state=present
+      with_items:
+      - iptables-services
+      when: ansible_distribution == 'CentOS' or ansible_distribution == 'Red Hat Enterprise Linux'
```
tiewei: use `ansible_os_family` instead

roles/etcd/tasks/main.yml
```diff
@@ -1,61 +1,74 @@
...
+
+    - name: Install iptables-persistent
+      apt: name={{item}} state=installed
+      with_items:
+          - iptables-persistent
+      when: ansible_distribution == 'Debian' or ansible_distribution == 'Ubuntu'
```
tiewei: ditto

roles/etcd/tasks/main.yml
```diff
@@ -1,61 +1,74 @@
...
+
+    - name: Set facts from role defaults for sharing with other hosts
+      set_fact:
+        etcd_node_name: "{{ etcd_node_name }}"
+        etcd_peer_advertise_urls: "{{ etcd_peer_advertise_urls }}"
+
+    - name: template etcd docker environment file
+      template:
+        src: etcd_env_file.j2
+        dest: "/etc/etcd_env"
```
tiewei: it might be better named differently or in different dir. e.g. `/etc/contiv/etcd.env` or `/etc/contiv-etcd.env`

roles/etcd/tasks/main.yml
```diff
@@ -1,61 +1,74 @@
...
+        - "{{ etcd_peer_port1 }}"
+
+    #  Save rules into /etc/sysconfig/iptables file for restoring rules on boot
+    - name: Save iptables
+      command: service iptables save
+      when: ansible_distribution == 'CentOS' or ansible_distribution == 'Red Hat Enterprise Linux'
+
+    - name: Set facts from role defaults for sharing with other hosts
+      set_fact:
+        etcd_node_name: "{{ etcd_node_name }}"
```
tiewei: would be helpful having `contiv` prefix, or just `contiv_etcd{0,3}`

roles/etcd/defaults/main.yml
```diff
@@ -1,14 +1,53 @@
 ---
 # role variable for the etcd service
 
+etcd_version: v2.3.8
+etcd_cluster_name: contiv
+
+etcd_proxy: "{{ etcd_proxies_group in groups
+                 and inventory_hostname in groups[etcd_proxies_group] }}"
```
tiewei, line 8: is there any chance `etcd_proxies_group in groups` and not `inventory_hostname in groups[etcd_proxies_group]` ?

roles/etcd/defaults/main.yml
```diff
@@ -1,14 +1,53 @@
...
+etcd_member: "{{ etcd_members_group in groups
+                 and inventory_hostname in groups[etcd_members_group] }}"
```
tiewei, line 10: ditto

roles/etcd/defaults/main.yml
```diff
@@ -1,14 +1,53 @@
...
+
+etcd_data_dir: /var/lib/etcd/data
+etcd_wal_dir: /var/lib/etcd/wal
+
 etcd_reset_state: false
+
+# Etcd has "advertise" urls for the other nodes to use if it's behind a proxy
+# Also, it needs to "listen" on IPs and ports for peers to talk to
+etcd_peer_advertise_interface: "{{ ansible_default_ipv4.interface }}"
+etcd_peer_advertise_address: "{{
```
tiewei, line 38: I think we should just retire `etcd_peer_advertise_interface` and use list of ips for  `etcd_peer_advertise_address`, this way if the default interface has more than one ips, it will have problems

vhosakot (approved):

roles/etcd/defaults/main.yml
chrisplo: oh how about etcd_proxies_group in group_names

roles/etcd/defaults/main.yml
chrisplo: inv can set `etcd_peer_advertise_address` in that scenario, this makes it easy for VMs (like molecule) where the interface is the same across all the VMs, otherwise that ugly ip fetch is pushed up higher level

roles/etcd/defaults/main.yml
tiewei: yeah, that's better

roles/etcd/defaults/main.yml
tiewei: i think molecule is special cases and usually it should just advertise  `ansible_default_ipv4.address`, this also avoids overridded `etcd_peer_advertise_address` doesn't match `etcd_peer_advertise_interface`,  also it's a little strange to say a service listen on a interface, but listen on a ip:port

roles/etcd/tasks/main.yml
chrisplo: could you elaborate, I'm not sure what you are asking or the benefit of namespacing here, it's not something end user interacts with

roles/etcd/defaults/main.yml
chrisplo: etcd_proxies_group in groups was to validate i didn't get a bad dereference, otherwise I would have just checked `inventory_hostname in groups[etcd_proxies_group]` . . . if I could do |default(False) that would be cool but i didn't test that out

chrisplo:
having some trouble with final validation of etcd unit interaction with docker, restart works well but stop/start having some issue

roles/etcd/defaults/main.yml
chrisplo: default ipv4 in vbox is 10.0.2.15 which is not useful, have to target the second interface

chrisplo: also contiv provides a control_interface, does not provide ip addresses

roles/etcd/tasks/main.yml
chrisplo: done

roles/etcd/tasks/main.yml
chrisplo: done

roles/etcd/tasks/main.yml
chrisplo: done

roles/etcd/defaults/main.yml
chrisplo: i tried out the etcd_proxies_group in group_names, group_names was not defined, maybe it's ansible 2.3

tiewei: hum, it's a feature since 1.9 ... . anyway, it's not a blocker

roles/etcd/defaults/main.yml
tiewei: `control_interface` is one of things i think we need to remove from etcd dependency i think, it was used in many places but most places are used to get it's ip address like here. 
I think vbox is a special case we need to workaround by set `etcd_peer_advertise_address` and leave default as `ansible_default_ipv4.address`.
This is not a blocker for refactoring etcd role, but this is going to be helpful moving forward

roles/etcd/tasks/main.yml
tiewei: not when they don't have to debug it ... indicate this is etcd for contiv instead of other general purpose etcd

roles/etcd/defaults/main.yml
chrisplo: good ideas, just there's more than enough scope with this change

chrisplo:
Fixing an issue with docker stop/start not triggering v2plugin start after etcd start

tiewei (approved):

tiewei:
LGTM

chrisplo:
the last commit is a cherry-pick of #380, if that lands separately we can remove that commit from this PR

dseevr (approved):

chrisplo:
I pulled the commit that was equivalent to #380 since that merged separately

unclejack (approved):
LGTM

chrisplo:
This helps fix https://github.com/contiv/netplugin/issues/963

PR #379, opened 2017-10-25
iptables-services package is no longer used

chrisplo:
roles/auth_proxy/files/contivRule.sh will save and restore iptables
rules, which means iptables-services is not longer needed, which makes
some installations simpler (default RHEL, or using firewalld)

Signed-off-by: Chris Plock <chrisplo@cisco.com>


tiewei (approved):

g1rana:
LGTM

tiewei:
LGTM

PR #384, opened 2017-12-14
Update ansible roles for netplugin, netmaster updated CLI

tiewei:
We refactored netplugin and netmaster CLI, and these changes are
required in order to make ansible able to deploy updated services.

Signed-off-by: Wei Tie <wtie@cisco.com>


tiewei:
this may not that important anymore

PR #386, opened 2018-02-12
roles/{contiv,etcd}: retry pulling image & plugin

unclejack:
This should help avoid issues with interrupted downloads when dealing with network connectivity problems.

This should help work around https://github.com/contiv/install/issues/340


tiewei (approved):

chrisplo (approved):
A bit blunt, but should help

dseevr (approved):

